PR #4, opened 2019-02-17
First draft of R2

strengejacke:
Support for first simple models, and maybe a good draft to think about design choices (here especially the `type`-argument).


strengejacke:
Maybe we can add aliases for them, but I would prefer to keep short-named alternatives of the functions as well... Having "direct measure" like `r2_nagelkerke()` or `performance_r2_nagelkerke()` sounds good!

DominiqueMakowski:
Keeping both a short name and a long name is indeed a solution for both accessibility (finding package functionalities for newcomers) and efficiency (for programmers)!

strengejacke:
Shall we add more functionality for this PR, or merge this one and then add step-by-step further models?

DominiqueMakowski:
let's merge it!

PR #6, opened 2019-02-20
add r2() support for Stan and mixed models

strengejacke:


strengejacke:
This mostly copies the code from sjstats, with minor modifications. There are some things to do, which are especially design choices: How to deal with Bayesian R2, that also returns a standard error? Should we add a class attribute for printing? Should the returned vector be names? Etc.

PR #8, opened 2019-02-23
Reorganisation

DominiqueMakowski:
- [x] Passing tests
- [ ] Fix model_performance.merMod for merMod_2 and merMod_3
- [ ] Add `model_performance()` methods for other models
- [ ] Improve docs
- [ ] Improve testing coverage
- [x] Remove `sjstats_r2.R` file


DominiqueMakowski:
So I've reorganised the things a bit, splitted the functions into different files. R2s are divided by their algorithm, and the `r2()` function is a wrapper returning the appropriate R2 metric. On top of that, `model_performance()` returns the R2(s) and other relevant metrics.

Tell me what you think

DominiqueMakowski:
Nice trick for the `r2_all()`!

I wonder the usage of such function though, in what cases aside for comparison studies would some need all R2s at once... but anyway we can add it down the road.

Before merging:
- Can we remove the sjstats_r2 file?
- What to do with `r2_linmix`?

strengejacke:
Yes, the sjstats_r2 file can be removes, just a placeholder. 
`r2_linmix()` requires the `icc()` function to work. So if we want to implement r2() for lme, we need to do some work. I can do this once this PR has been merged, so I have a "proper" base to develop further.

PR #9, opened 2019-02-26
Minor revisions for r2()

strengejacke:
- Make sure `r2_bayes()` returns only one value, not the posterior
- add function that returns the posterior, for summary-function


strengejacke:
I'm not sure whether it's a bug or intended that brms ignores the `re.form` argument (and does not work with `re_formula`, either). I have filed an issue (https://github.com/paul-buerkner/brms/issues/610) and await Paul's response.

Issue #12, opened 2019-03-01
ICC


strengejacke:
Currently, `icc()` works for the packages *lme4*, * glmmTMB*  and *rstanarm*. *brms* calculates the ICC in a different way, based on a variance-decomposition. For non-Gaussian models, this approach is also recommended by Stan team members. For non-Bayesian models (lme4/glmmTMB), the ICC for non-Gaussian models is based on some approximations of the variances (which is not perfect, but apparently the current state of the art).

Questions:

1) Do we want rstanarm models behave like lme4 or brms? The latter would mean that non-Gaussian models fitted with rstanarm would return a more "Bayesian correct" ICC.

2) For brms, we can't use the variance-calculation we use for the `r2_nagakawa()`. We could only calculate a comparable ICC for brms models when they are Gaussian and only have one random intercept. Do we want to cover this "edge case" by implementing extra code particularly for these types of models, or do we always want the variance decomposition for brms?

Let me give a comparison, only for the same brms-model

```r
model4 <- brms::brm(Petal.Length ~ Petal.Width + (1 | Species), data = iris)

# Variance decomposition
performance::icc(model4)
#> $ICC_decomposed
#> [1] 0.7477364
#> 
#> $CI
#>      2.5%     97.5% 
#> 0.8543512 0.6156847 

# ICC only for linear models, with only random intercept
sjstats::icc(model4)
#> ## Species
#>           ICC: 0.94  HDI 89%: [0.86  1.00]
```

If you would use the current `icc()` for a mixed model from `rstanarm`, you get the ICC based on Nagakawa, as it would be returned for lme4 etc. as well:

```r
performance::icc(model2)
#> $ICC_adjusted
#> [1] 0.8786005
#> 
#> $ICC_conditional
#> [1] 0.5603696
```

So, method 1) (Variance Decomposition) gives an ICC of .75, method 2) (simple ICC for linear models with random intercept only) gives ans ICC of .94 while method 3) (Nagakawa ICC) returns .88 and .56.

We could use method 1) or 3) for brms linear models with random intercept only, where method 1) is below (or between those two ICC values) and method 3) is above the values of method 2)...

3) What values should be returned? We have the CI for brms-models, but only one "ICC" (which is not exactly an ICC, but something comparable, the variance decomposition in a Bayesian way...)

4) How do we call what `icc()` returns for brms models? Still ICC, and explain this in the docs of `icc()`?

strengejacke:
[...]

PR #13, opened 2019-03-01
Close #3

### LINKED ISSUE #3: Marginal vs. conditional R2 for Bayesian models
It would be a great feature to have something for that. Options are:

- Expand `R2_nakagawa` to work with Bayesian models
- Adapt `R2 MLM` to work with Bayesian models
- Create a function based on `rstanarm::bayes_R2`

For the last option, [this thread](https://discourse.mc-stan.org/t/marginal-and-conditional-r2-for-stans-glmms/4557/5) could be relevant:

> bayes_R2(model, re.form = NULL) vs bayes_R2(model, re.form = NA) is fine, but it shouldn’t be used to “test” a null hypothesis that the group-specific intercept and coefficients are irrelevant.

strengejacke:
- implements the `r2_bayes()`, working with the forthcoming update
- implements `icc()`-methods for lme4, glmmTMB, rstanarm and brms


strengejacke:
Please also see #12 in this context.

DominiqueMakowski:
Mmh, I think my knowledge of ICC is currently too low for any enlighted opinion... I will try to read and learn tomorrow. But in the meantime, trust the force 👽 and do how you feel ☺️

strengejacke:
Ok. At this point, this PR won't break anything nor push something existing in _performance_ into a completely different direction, so it's safe to merge, I would say. And we can later still change or improve the implementation of `icc()`

PR #16, opened 2019-03-03
update

DominiqueMakowski:


DominiqueMakowski:
I have remarked that: `performance::r2(circus::download_model("merMod_1"))` errors, is it expected?

strengejacke:
The code (from sjstats) used the old element-names from `model_family()`, not the revised from `insight::model_info()`. Should be fixed now.

PR #18, opened 2019-03-04
Add methods for internal consistency

strengejacke:
Short question: Do we want one help-page for each method, or "collect" methods like I did in sjstats? (i.e. rmse, mse etc. under "model quality", reliability, cronbach under "internal consistency"?)


DominiqueMakowski:
As you can guess, I have a tendency to prefer one help-page / method. Although a bit more tedious to maintain and write, I feel like it's easier for the users to look up the information... but I might be wrong

strengejacke:
I'm in favour of either way, so we can keep it as it currently is.

strengejacke:
We need to push bayestestR on CRAN for this package to be released ;-)

R/mean_iic.R
```diff
@@ -0,0 +1,69 @@
...
+#'
+#' @examples
+#' data(mtcars)
+#' x <- mtcars[, c("cyl", "gear", "carb", "hp")]
+#' mean_iic(x)
+#'
+#'
+#' @importFrom stats cor na.omit
+#' @export
+mean_iic <- function(x, method = c("pearson", "spearman", "kendall")) {
```
DominiqueMakowski: I wonder if for item-related methods, it wouldn't be nice to have a consistent prefix like item_difficulty, item_iic (altough *technically* redundant) etc. so it gives this feeling of organisation?

strengejacke: yeah, I was thinking about something in this direction as well...  Would you apply this to all methods? I.e. 
- item_reliabitily
- item_cronbach
- item_split_half
- item_iic
- item_difficulty

R/cronbach.R
```diff
@@ -0,0 +1,39 @@
...
+#'    and everything > 0.8 is good or excellent.
+#'
+#' @examples
+#' data(mtcars)
+#' x <- mtcars[, c("cyl", "gear", "carb", "hp")]
+#' cronbach(x)
+#'
+#' @importFrom stats var na.omit
+#' @export
+cronbach <- function(x) {
```
DominiqueMakowski: Would `cronbach_alpha()` be more transparent (or is Cronbach's only legacy this metric ^^?)

R/reliability.R
```diff
@@ -0,0 +1,86 @@
...
+#'    reversed answer categories regarding positive and negative poles).
+#'
+#' @examples
+#' data(mtcars)
+#' x <- mtcars[, c("cyl", "gear", "carb", "hp")]
+#' reliability(x)
+#'
+#' @importFrom stats cor var na.omit
+#' @export
+reliability <- function(x, standardize = FALSE, digits = 3) {
```
DominiqueMakowski: same as above, maybe `item_reliability()` would be more specific and explicit?

DominiqueMakowski:
- item_reliabitily
- item_cronbach
- item_split_half
- item_iic
- item_difficulty

Yes for all, I only have doubts for item_cronbach, which seems to be known "as is" well enough for having its own unique method?

R/cronbach.R
strengejacke: > or is Cronbach's only legacy this metric

I don't know, you are the psychologist! :-)
In times of auto-completion, I would say we can use `cronbachs_alpha()`.

strengejacke:
- item_reliabitily
- item_split_half
- item_difficulty

I think these are ok.

- `item_cronbach()`  should be `cronbachs_alpha()`.

`item_iic()` - since one of the i's stands for "item", i'm not quite sure how to name this one. Following styles come into my mind:
1) `item_iic()`
2) `item_mic()` (for mean)
3) `item_meancor()` (sounds strange...?)

R/cronbach.R
DominiqueMakowski: It can also increase its google's findability 👀    (sometimes I wonder if my psychologist-part is getting slowly taken over by a marketing/salesman part 😱 )

DominiqueMakowski:
That last one is tricky indeed 😕 it's complicated to fit many words into a transparent unique one... `item_intercor`?

strengejacke:
Should we prefix rmse, mse, rse and error_rate as well? The best thing that comes to my mind is a `model_` prefix, though this is not very specific.

DominiqueMakowski:
I think RMSE and MSE are quite known, I would say keep them like this. Consistently with that, the same would apply for RSE. As for error_rate... we can also leave it like that for now.

PR #80, opened 2019-08-14
R2 bayes changes

DominiqueMakowski:
Minor R2 bayes adjustements for condsistency and usage by report

- store all point estimates for R2_bayes
- pass on attributes to model_performance
- Change name of r2_loo_adjusted to r2_adjusted to follow consistency of SE etc.
- still think SE should be renamed to match what is actually computed (SD or MAD)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: performance
 Type: Package
 Title: Assessment of Regression Models Performance
-Version: 0.3.0.9000
+Version: 0.3.1.9000
 Date: 2019-08-05
 Authors@R: c(
 	person("Daniel", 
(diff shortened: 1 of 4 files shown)


strengejacke:
> still think SE should be renamed to match what is actually computed (SD or MAD)

Isn't the SE just the SD of the sampling distrbution?

> The standard deviation of an estimate is called the standard error.
(https://support.minitab.com/en-us/minitab/18/help-and-how-to/modeling-statistics/regression/supporting-topics/regression-models/what-is-the-standard-error-of-the-coefficient/)

DominiqueMakowski:
> Isn't the SE just the SD of the sampling distribution? 

I agree, but it's just that SE sounds very frequentist, and I think it might confuse many people to see SE in the output of the Bayesian R2 (and here it is the really raw SD of a distribution and not an inferred/estimated measure of uncertainty)

DominiqueMakowski:
But then having consistent output makes also sense, hence I kept SE but passed point-estimates and dispersion measures as attributes :)

strengejacke:
rstanarm also returns `sd` as column name... brms returns "estimation error". our `SE` is something in between that, maybe we can actually think of renaming it to `SD`, though it likely will break code. Or we point to it in the docs.

PR #82, opened 2019-08-19
Ultimate outliers detection: (fixes and improvements to check_outliers)

DominiqueMakowski:
This PR aims at fixing, improving and extending the check_outliers methods

As I need to implement the method for dataframes, I read a bit more in-depth about outliers detection, including some papers about Mahalabonis and all, resulting in some fixes and corrections in the default thresholds. 

Currently didn't touch much the .default method, and only updated the details section of the docs. Also, the .data.frame methods works but the legacy print method is not compatible (hence it errors).

Basically, I allowed for multiple methods at a time and implemented the composite outlier score (the average of the classifications, i.e., the probability for an observation of being classified as an outlier by one method).

The print and the plotting will need to be updated, but I am not sure how. I will make a candidate update for the model method


strengejacke:
38dia2

DominiqueMakowski:
I am quite satisfied with the current implementation, waiting for your feedback.

## Regarding the plot method

In the case of a single method specified, we could keep the previous method, and in the case of a multiple methods, add something like this:

``` r
# Plot test
library(performance)
mt1 <- mtcars[, c(1, 3, 4)]
mt2 <- rbind(mt1, data.frame(mpg = c(37, 40), disp = c(300, 400), hp = c(110, 120)))
model <- lm(disp ~ mpg + hp, data = mt2)

x <- check_outliers(model, method = "all")

# Data plot ---------------
library(parameters)
#> 
#> Attaching package: 'parameters'
#> The following object is masked from 'package:performance':
#> 
#>     principal_components

data_plot_outliers <- function(x){
  data <- attributes(x)$data
  row.names(data) <- data$Obs

  # Extract distances
  d <- data[grepl("Distance_", names(data))]
  d <- parameters::normalize(d)

  d_long <- stats::reshape(
    d,
    direction = "long",
    varying = list(names(d)),
    sep = "_",
    v.names = "Distance",
    timevar = "Method",
    times = names(d)
  )
  d_long$Obs <- as.factor(d_long$id)
  row.names(d_long) <- d_long$id <- NULL
  d_long$Method <- gsub("Distance_", "", d_long$Method)
  d_long
}


# Plot ---------------
library(tidyverse)
#> Warning: package 'dplyr' was built under R version 3.6.1
library(see)

x <- check_outliers(model, method = "all")


x %>%
  data_plot_outliers() %>%
  ggplot(aes(x = Obs, y = Distance, fill = Method, group = Method)) +
  # geom_vline(xintercept = as.character(c(1, 2))) +
  geom_bar(position = "dodge", stat = "identity") +
  scale_fill_viridis_d() +
  theme_modern()  +
  theme(axis.text.x = element_text(colour = ifelse(as.numeric(x) >= 0.5, "red", "darkgrey")),
        panel.grid.major.x = element_line(linetype = "dashed", colour = ifelse(as.numeric(x) >= 0.5, "red", "lightgrey")))
```

image


<sup>Created on 2019-08-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


We can clearly see the problematic observations that have more overall "visual mass" to them

strengejacke:
looks good! shall I implement a related plot-method in see?

strengejacke:
Not sure how to refer to `x` in the last two lines (`theme(...)`).

strengejacke:
ok, got it

PR #87, opened 2019-09-15
remove principal_components to avoid conflict

DominiqueMakowski:


strengejacke:
Do you want to remove it directly before deprecating it?

DominiqueMakowski:
I think it should be okay to just remove it, but as you want

strengejacke:
Just saw in sjPlot I use `performance:: principal_components()`. So I have to update sjPlot before performance.

DominiqueMakowski:
oops 🙊

PR #100, opened 2020-02-04
few more models supported in model_performance using default method

IndrajeetPatil:


tests/testthat/test-model_performance-various.R
```diff
@@ -13,3 +18,35 @@ if (require("testthat") && require("performance") && require("insight") && requi
...
+  res1 <- DirichReg(Y ~ depth + I(depth^2), ALake)
+
+  test_that("model_performance (Dirichlet regression)", {
+    expect_equal(
+      performance::model_performance(res1),
+      structure(
+        list(
+          AIC = -199.993722776335,
+          BIC = -185.021667961168,
+          NA. = NA,
```
strengejacke: Why does it return element `NA.` here?

IndrajeetPatil: As far as I can see, this has to do with the following warning printed during execution:

> 'r2()' does not support models of class 'DirichletRegModel'.

So, instead, the function just introduces `NA`s. Is that correct?
```r
>     performance::r2(res1) 
'r2()' does not support models of class 'DirichletRegModel'.
[1] NA
```

strengejacke: Yes, I just saw that, too. I've implemented r2() for DirichletReg (which calls Nagelkerke's R2).

strengejacke:
You could merge the master into this branch, and then fix the test, then it should be OK to merge.

strengejacke:
Thanks!

IndrajeetPatil:
@strengejacke Can you please add me to the `DESCRIPTION` file? 

I realized now that I was supposed to do it but forgot to do it. :)

strengejacke:
Done :-)

PR #216, opened 2021-03-18
more edits to the paper

IndrajeetPatil:


paper/paper.bib
```diff
@@ -43,14 +43,14 @@ @Manual{robinson_broom_2020
 }
 
 @Article{ludecke2020see,
-  title = {{see}: Visualisation Toolbox for 'easystats' and Extra Geoms, Themes and Color Palettes for 'ggplot2'},
```
strengejacke, line 46: I think we need to preserve the curly braces, and furthermore even add more of them to the title - everywhere, where we need to force to preserve upper/lower case. Thus, package names, "easystats" or "R" need to be surrounded by `{}`.

PR #247, opened 2021-03-31
suppress printout in model_performance.gamlss

vincentarelbundock:
`gamlss` inherits from `gam`, and `summary.gamlss` automatically prints out a summary to the console when called. This captures the printout to ensure that `model_performance` is silent when assigned.

```{r}
library(gamlss)
library(performance)

dat <- rgamma(100, shape=1, scale=10)
mod <- gamlss(dat ~ 1, family = GA, trace = FALSE)
x <- model_performance(mod)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -27,7 +27,12 @@ Authors@R:
              family = "Waggoner",
              role = c("aut", "ctb"),
              email = "philip.waggoner@gmail.com",
-             comment = c(ORCID = "0000-0002-7825-7573")))
+             comment = c(ORCID = "0000-0002-7825-7573")),
+      person(given = "Vincent",
+             family = "Arel-Bundock", 
+             email = "vincent.arel-bundock@umontreal.ca", 
+             role = "ctb",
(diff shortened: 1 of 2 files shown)


strengejacke:
Thanks for tracking this down. Can you add the namespace `utils::` and add a roxygen `@importFrom` if necessary?

strengejacke:
This PR closes #247

vincentarelbundock:
For some reason I thought `utils` was on the base R search path, such that we didn't have to specify the namespace (as opposed to `tools` for example. ¯\_(ツ)_/¯

I updated the PR and force-pushed for clean history.

I don't think `@importFrom` is needed, but happy to add it if you want.

strengejacke:
> I thought utils was on the base R search path, such that we didn't have to specify the namespace

Yeah, but it's a personal habit that I add namespace for every function except *base* ;-)
PR looks good, thanks!

strengejacke:
You might want to add yourself as contributor to the "DESCRIPTION".

vincentarelbundock:
Done, thanks!

PR #257, opened 2021-04-04
Add test_lrt as an alias to test_likelihoodratio

bwiernik:
I like that the `test_*` naming convention was adopted. I think that having an alias using "lrt" in addition to the longer name would be helpful for discoverability and to reduce typing.


DominiqueMakowski:
Looks good thanks ☺️

Issue #256, opened 2021-04-04
CIs for frequentist regression R2 and sigma


bwiernik:
Would it be possible to add functions for computing confidence intervals on regression model sigma and R2 values? These are notably absent from the `stats` `confint.lm` and `confint.glm` methods. These would be nice to optionally include in `r2()` or `performance()` output.

I've included functions to compute these for `lm` below. Is there interest in building these out further?

<details>

```r
confint_sigma <- function(x, conf.level = .95, ...) {
  alpha <- 1 - conf.level
  dev <- deviance(x)
  n <- nobs(x)
  lb <-  dev / qchisq(1 - alpha / 2, n)
  ub <-  dev / qchisq(alpha / 2, n)
  return(
    c(sigma.lower = sqrt(lb), sigma.upper = sqrt(ub))
  )
}

confint_Rsq <- function(x, conf.level = .95, ...) {
  alpha <- 1 - conf.level
  n <- nobs(x)
  df.int <- if (attr(x$terms, "intercept")) 1L else 0L
  p <- x$rank - df.int
  R2_obs <- summary(x)$r.squared
  lb <- uniroot(pRsq,
                c(.00001, .99999),
                alpha = 1 - alpha / 2,
                n = n,
                p = p,
                R2_obs = R2_obs)$root
  ub <- uniroot(pRsq,
                c(.00001, .99999),
                alpha = alpha / 2,
                n = n,
                p = p,
                R2_obs = R2_obs)$root
  return(
    c(r.squared.lower = lb, r.squared.upper = ub)
  )
}

dRsq <- function(K1, R2_pop, R2_obs, p, nobs) {
  NCP <- R2_pop / (1 - R2_pop)
  F1_obs <- ((nobs - p - 1) / p) * (R2_obs / (1 - R2_obs))
  exp(
    log(
      pf(q = F1_obs,
         df1 = p,
         df2 = (nobs - p - 1),
         ncp = NCP * K1,
         lower.tail = FALSE)
    ) +
      dchisq(x = K1, df = (nobs - 1), log = TRUE)
  )
}

pRsq <- function(R2_pop, R2_obs, p, nobs, alpha = 1) {
  a1 <- 1 - alpha
  # This approach avoids undersampling the area of the chi-squared
  # distribution that actually has any density
  integrals <-
    mapply(function(i, j, ...) {
      dots <- list(...)
      integrate(dRsq,
                i, j,
                R2_pop = dots$R2_pop,
[...]

PR #259, opened 2021-04-05
add R2 CI

strengejacke:


strengejacke:
closes #256

strengejacke:
Looks good, @bwiernik @DominiqueMakowski ?

``` r
library(performance)
m <- lm(formula = mpg ~ wt + qsec + am, data = mtcars)
performance:::confint_r2(m)
#>          R2    CI_low  CI_high
#> 1 0.8496636 0.6820641 0.916607
```

<sup>Created on 2021-04-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

strengejacke:
Any suggestions how to implement this? Should we add an additional argument (like `ci`, that defaults to `NULL` / `NA`, and if given, includes the CI?)

strengejacke:
@bwiernik Any references for this approach that we could add to the docs?

bwiernik:
Based on Shieh, G. (2006). Exact interval estimation, power calculation, and sample size determination in normal correlation analysis. Psychometrika, 71(3), 529. https://doi.org/10.1007/s11336-04-1221-6

strengejacke:
for some models, it fails. So we should wrap this function into `tryCatch()`:

``` r
library(performance)
m <- lm(formula = mpg ~ wt + qsec + am, data = mtcars)
performance:::.confint_r2(m)
#> $R2
#>        R2    CI_low   CI_high 
#> 0.8496636 0.6820641 0.9166070 
#> 
#> $R2_adjusted
#> adjusted R2      CI_low     CI_high 
#>   0.8335561   0.6516598   0.9072481


model <- glm(vs ~ wt + mpg, data = mtcars, family = "binomial")
performance:::.confint_r2(model)
#> $R2_Tjur
#> Tjur's R2    CI_low   CI_high 
#> 0.4776926 0.1728083 0.6836999


# example from ?glm
counts <- c(18,17,15,20,10,20,25,13,12)
outcome <- gl(3,1,9)
treatment <- gl(3,3)
d <- data.frame(treatment, outcome, counts)
glm.D93 <- glm(counts ~ outcome + treatment, family = poisson(), data = d)
performance:::.confint_r2(glm.D93)
#> Error in stats::uniroot(.pRsq, c(1e-05, 0.99999), R2_obs = as.vector(rsq), : f() values at end points not of opposite sign
```

<sup>Created on 2021-04-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

bwiernik:
I've only thought about this in the context of linear models, so I'm not sure about performance for GLMs, especially not for various flavors of R2.

bwiernik:
Perhaps we should add a `method` argument and include bootstrap as an option. We could limit methods to bootstrap unless a solid analytic method is available.

DominiqueMakowski:
> Should we add an additional argument (like ci, that defaults to NULL / NA, and if given, includes the CI?)

I think that models where an analytical method is straightforward we can return it by default. Better to have more information by default (especially popular concepts like CI) than not

bwiernik:
@strengejacke Just checking on the status of this? What are the issues still to be resolved? I think it would be useful to merge in the option for CIs for R2/sigma on `lm` and Gaussian `glm` models. We can check on other forms of GLMs and multilevel models in the future. 

For Bayesian models, this is already all handled by the posterior.

mattansb:
Just to note that in `effectsize`....

```R
m <- lm(mpg ~ ., mtcars)

summary(m)
#> 
#> Call:
#> lm(formula = mpg ~ ., data = mtcars)
#> 
#> Residuals:
#>     Min      1Q  Median      3Q     Max 
#> -3.4506 -1.6044 -0.1196  1.2193  4.6271 
#> 
#> Coefficients:
#>  Estimate Std. Error t value Pr(>|t|)  
#> (Intercept) 12.30337   18.71788   0.657   0.5181  
#> cyl         -0.11144    1.04502  -0.107   0.9161  
#> disp         0.01334    0.01786   0.747   0.4635  
#> hp          -0.02148    0.02177  -0.987   0.3350  
#> drat         0.78711    1.63537   0.481   0.6353  
#> wt          -3.71530    1.89441  -1.961   0.0633 .
#> qsec         0.82104    0.73084   1.123   0.2739  
#> vs           0.31776    2.10451   0.151   0.8814  
#> am           2.52023    2.05665   1.225   0.2340  
#> gear         0.65541    1.49326   0.439   0.6652  
#> carb        -0.19942    0.82875  -0.241   0.8122  
#> ---
#> Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
#> 
#> Residual standard error: 2.65 on 21 degrees of freedom
#> Multiple R-squared:  0.869,	Adjusted R-squared:  0.8066 
#> F-statistic: 13.93 on 10 and 21 DF,  p-value: 3.793e-07 # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< Here

performance::r2(m)
#> # R2 for Linear Regression
#>        R2: 0.869
#>   adj. R2: 0.807

# Same estimates + CIs
effectsize::F_to_eta2(13.93, 10, 21) |> print(digits = 3)
#> Eta2 (partial) |         95% CI
#> -------------------------------
#> 0.869          | [0.743, 1.000]
#> 
#> - One-sided CIs: upper bound fixed at (1).
effectsize::F_to_eta2_adj(13.93, 10, 21) |> print(digits = 3)
#> Epsilon2 (partial) |         95% CI
#> -----------------------------------
#> 0.807              | [0.609, 1.000]
#> 
#> - One-sided CIs: upper bound fixed at (1).
```

strengejacke:
Running checks now

bwiernik:
It doesn't look like the CI is actually added to the output?

Issue #268, opened 2021-04-15
check_model error


delucalab:
Hope you are staying safe! I am new to all this so forgive me if this is a naive question. I am using the check_model function on a model I built with lmer in the lme4 package. Some of my predictors are dummy coded variables (0, 1) which I set as factors. I then received the following error:

_**Error in 1:label.n : NA/NaN argument**_

When I don't set the predictors as factors I no longer receive the error. However, in another data set with a mix of dummy coded variables (0,1,2) and (0,1) I don't get this error if they are all set as factors. 

Therefore, I am a bit confused about the error and wanted to see if you had any insight?

Warm Regards,
Alex

strengejacke:
Do you have a small reproducible example?

delucalab:
Attached are the datasets and associated code to reproduce the errors/non-errors. Let me know if you need anything else. Thanks for your help!
[Dataset 1.csv](https://github.com/easystats/performance/files/6318715/Dataset.1.csv)
[Dataset 2.csv](https://github.com/easystats/performance/files/6318716/Dataset.2.csv)
[Reproducible Example.txt](https://github.com/easystats/performance/files/6318717/Reproducible.Example.txt)

bwiernik:
(This is going to be in the new cooks distance plot, label.n controls how many points get text annotations.)

delucalab:
I looked further and I noticed that I only get the error (Error in 1:label.n : NA/NaN argument) if the input model does not have any continuous predictors and I make the dummy variable factors. However, if I do not use the as.factor command and solely have discrete predictors or I add a continuous predictor while still using the as.factor command, I know longer get the error.

bwiernik:
Can you please give a small reproducible example we can run?

delucalab:
I'm attaching a dataset file (Dataset 2.csv) and the code (Reproducible Example.txt) to reproduce the situations when errors arise and don't. The text file should indicate which set of commands should produce the error and which won't (including what is different between the data). Let me know if you need anything else or if I'm not providing the right format or something (this is all new to me).
[Dataset 2.csv](https://github.com/easystats/performance/files/6319183/Dataset.2.csv)
[Reproducible Example.txt](https://github.com/easystats/performance/files/6319184/Reproducible.Example.txt)

delucalab:
[...]

Issue #186, opened 2021-01-11
test_vuong for mixed models


DominiqueMakowski:
As far as I understand, two things are required for mixed models to work with that test:

- **Individual loglikelihoods**: This is very strange. Because the .glm method seems to work natively for .glmer (https://github.com/easystats/insight/blob/master/R/loglikelihood.R#L159), but the .lm method doesn't for lmer. Moreover, I simply cannot find the function definition of logLik to see what it does! I've looked in base R and in lme4, don't know where it's hidden. 

``` r
x <- lme4::lmer(Sepal.Length ~ Sepal.Width + (1|Species), data = iris)
stats::logLik(x)
#> 'log Lik.' -97.31807 (df=4)
```

<sup>Created on 2021-01-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This shouldn't be too hard to reimplement to get the non-summed LLs, once we find how logLik is currently implemented.

- **estfun**: So in vuong test one of the two dependency currently required is `sandwich::estfun()`, which computes *Empirical Estimating Functions* of models. From my (overly)quick reading, it seems to be some sort of empirical approximation of something loosely related to loglik that is especially useful in some cases (e.g., for complex models). 🤷 

In any case, it seems that for most of cases, `estfun` are equivalent to the residuals multiplied by the models' matrix:

``` r
x <- lm(mpg ~ disp * wt, data=mtcars)

ref <- sandwich::estfun(x)
head(ref)
#>                   (Intercept)       disp         wt    disp:wt
#> Mazda RX4         -1.95292297 -312.46768 -5.1166582 -818.66531
#> Mazda RX4 Wag     -0.77410571 -123.85691 -2.2255539 -356.08863
#> Datsun 710        -3.05824938 -330.29093 -7.0951386 -766.27497
#> Hornet 4 Drive     3.03272404  782.44280  9.7502078 2515.55361
#> Hornet Sportabout  2.75608748  992.19149  9.4809409 3413.13873
#> Valiant            0.06097151   13.71859  0.2109614   47.46632


rez <- insight::get_residuals(x) * model.matrix(x)
head(rez == ref)
#>                   (Intercept) disp   wt disp:wt
#> Mazda RX4                TRUE TRUE TRUE    TRUE
#> Mazda RX4 Wag            TRUE TRUE TRUE    TRUE
#> Datsun 710               TRUE TRUE TRUE    TRUE
[...]

PR #269, opened 2021-04-15
Use model.matrix() in check_outliers instead of get_predictors()

bwiernik:
ncol(data) needs to refer to the number for terms in the model function. The current approach is incorrect because it (1) doesn't separate factor variables into levels and (2) doesn't include the intercept.

Closes https://github.com/easystats/performance/issues/268


R/check_outliers.R
```diff
@@ -184,8 +184,7 @@ check_outliers.default <- function(x, method = c("cook", "pareto"), threshold =
   method <- match.arg(method, c("zscore", "iqr", "cook", "pareto", "mahalanobis", "robust", "mcd", "ics", "optics", "iforest", "lof"), several.ok = TRUE)
 
   # Remove non-numerics
-  data <- insight::get_predictors(x)
-  data <- data[, sapply(data, is.numeric), drop = FALSE]
+  data <- model.matrix(x)
```
strengejacke: please use `stats::model.matrix()` (and add `@importFrom stats model.matrix`).

The only problem I see is that `model.matrix` is less stable than `get_predictors()`. E.g., `nlme::lme()` has no model.matrix method:

``` r
library(nlme)
fm1 <- lme(distance ~ age, data = Orthodont) # random is ~ age
model.matrix(fm1)
#> Error in eval(predvars, data, env): object 'distance' not found
```

<sup>Created on 2021-04-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

@DominiqueMakowski Maybe we should also think about implementing a more robust `model.matrix()` in _insight_?

strengejacke (changes requested):

R/check_outliers.R
bwiernik: That last point would be a good approach.

DominiqueMakowski: Yes, there are multiple reasons for that (e.g., [1](https://github.com/easystats/insight/issues/316#issuecomment-793537395) and [2](https://github.com/easystats/performance/issues/186)). We can call it `get_modelmatrix` or `model_matrix`?

bwiernik: `get_modelmatrix` `get_model_matrix` would be more consistent.

bwiernik: > @importFrom stats model.matrix)

If a function is called with its namespace `stats::model.matrix`, it isn't necessary to also import it. Those are redundant.

strengejacke: > If a function is called with its namespace stats::model.matrix, it isn't necessary to also import it. Those are redundant.

Why? Else, those won't be updated/listed in the NAMESPACE file.

bwiernik: They don't need to be in the NAMESPACE file if you namespace them when you call them. That is one of the major advantages of namespacing when you call them. Also helps avoid namespace collisions.

strengejacke: Ok, so you don't need to tell R that there is a selective import? Or is this only true for the "core" packages (stats, utils, ...)?

strengejacke: I just read this chapter: https://r-pkgs.org/namespace.html
and as far as I understood, it indeed seems like you said. In this case, I would vote for using namespace in function calls, and remove the `@importFrom`.

bwiernik: Yeah, you only need to importFrom if you want to use the function without the namespace (eg, ggplot2 functions or some stats functions might be worth importing)

PR #306, opened 2021-06-04
use check if installed; style

IndrajeetPatil:


IndrajeetPatil:
@DominiqueMakowski Do you know what is wrong with `check_distribution` here?

There was a `sysdata.rda` file, which I deleted and then there was error that there is no `classify_distribution` function.

So added it as an alias for `check_distribution`  and now those examples are failing.

Not sure what's going on here.

IndrajeetPatil:
Never mind, I just reverted those changes.

PR #328, opened 2021-07-11
Minor Fix: binned_residuals documentation

Demetrio92:
Documentation to `binned_residuals` tells us that it returns a data frame, but the default action is `print`. The code snippet afterwards incorrectly uses `plot` instead of print. This PR fixes it.


IndrajeetPatil:
Hi, the `.Rd` files are read-only. You can make this change here and then run `devtools::document()`:

https://github.com/easystats/performance/blob/7f5b941ca20e4123bc0e98feff08d33c566583b5/R/binned_residuals.R#L40

IndrajeetPatil:
I am including this change in https://github.com/easystats/performance/pull/333.
So closing in favor of that PR.

PR #347, opened 2021-07-30
Updating conclusion if alpha > 0.05.

andtheWings:
To my understanding, using Bartlett's test, you can't conclude that the variances are the same in absolute terms, you can only say there is insufficient evidence to claim otherwise. https://www.statology.org/bartletts-test/

Perhaps there is a better way to express this in the function output, but I think it's important not to mislead people who will be casually using the test without fully understanding its methodology.

### DIFF
--- a/R/check_homogeneity.R
+++ b/R/check_homogeneity.R
@@ -151,7 +151,7 @@ check_homogeneity.afex_aov <- function(x, method = "levene", ...) {
   } else if (p.val < 0.05) {
     insight::print_color(sprintf("Warning: Variances differ between groups (%s, p = %.3f).\n", method.string, p.val), "red")
   } else {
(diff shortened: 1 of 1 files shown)


DominiqueMakowski:
It's just a bit strange to have "OK: there is insufficient" 

it sounds like "yes well but actually no" ^^

why not "Ok: The hypothesis that the groups have different variances can be rejected"

andtheWings:
I agree it confusing wording. Perhaps it could say, `"OK: Accept the null hypothesis that the variances for each group are the same (%s, p = %.3f).\n"`

I think Section 4 in [this reference](https://ebn.bmj.com/content/16/2/36) explains the distinction well: "It is critical at this point to realise that we have not proven the null hypothesis to be correct...but have decided that the evidence is not robust enough to disprove it."

bwiernik:
Especially tricky given the power issues of this test. 

How about "Bartlett's test does not find clear evidence of heteroskedastocity (p = .35)."

bwiernik:
Thanks!

Issue #360, opened 2021-09-02
attaching performance after rstanarm breaks pp_check S3 dispatch


bwiernik:
If I `library(easystats)` after `library(rstanarm)`, it breaks S3 dispatch on the `pp_check()` function, sending the `stanreg` model to performance's `glm` method, rather than the `stanreg` method. We should see if we can fix that or otherwise check if a model is `stanreg` and then send to the `bayesplot()` method manually.

``` r
library(rstanarm)
#> Loading required package: Rcpp
#> This is rstanarm version 2.21.1
#> - See https://mc-stan.org/rstanarm/articles/priors for changes to default priors!
#> - Default priors may change, so it's safest to specify priors, even if equivalent to the defaults.
#> - For execution on a local, multicore CPU with excess RAM we recommend calling
#>   options(mc.cores = parallel::detectCores())
library(performance)
#> 
#> Attaching package: 'performance'
#> The following object is masked from 'package:rstanarm':
#> 
#>     pp_check
m <- stan_glm(mpg ~ disp, data = mtcars)
#> 
#> SAMPLING FOR MODEL 'continuous' NOW (CHAIN 1).
#> Chain 1: 
#> Chain 1: Gradient evaluation took 0 seconds
#> Chain 1: 1000 transitions using 10 leapfrog steps per transition would take 0 seconds.
#> Chain 1: Adjust your expectations accordingly!
#> Chain 1: 
#> Chain 1: 
#> Chain 1: Iteration:    1 / 2000 [  0%]  (Warmup)
#> Chain 1: Iteration:  200 / 2000 [ 10%]  (Warmup)
#> Chain 1: Iteration:  400 / 2000 [ 20%]  (Warmup)
#> Chain 1: Iteration:  600 / 2000 [ 30%]  (Warmup)
#> Chain 1: Iteration:  800 / 2000 [ 40%]  (Warmup)
#> Chain 1: Iteration: 1000 / 2000 [ 50%]  (Warmup)
#> Chain 1: Iteration: 1001 / 2000 [ 50%]  (Sampling)
#> Chain 1: Iteration: 1200 / 2000 [ 60%]  (Sampling)
#> Chain 1: Iteration: 1400 / 2000 [ 70%]  (Sampling)
#> Chain 1: Iteration: 1600 / 2000 [ 80%]  (Sampling)
#> Chain 1: Iteration: 1800 / 2000 [ 90%]  (Sampling)
[...]

PR #361, opened 2021-09-04
Better handle pp_check() S3 exporting

bwiernik:
Closes https://github.com/easystats/performance/issues/360


strengejacke:
As I commented [here](https://github.com/easystats/performance/issues/360#issuecomment-913111540), I'd suggest removing the onLoad-stuff and just just our own name for that function. @DominiqueMakowski any thoughts from your side?

bwiernik:
I don't think it's much of a problem to leave the S3 registration there

strengejacke:
ok, so this doesn't require higher a R version (at least 3.6)?

bwiernik:
No, it doesn't change our minimum version at all. 

I've found the clearerest and least disruptive approach I think. roxygen2 has a `@rawNamespace` command we can use to export the `if (getRversion() >= "3.6.0") { }` documentation. Then, we can simply say in the documentation that `On R 3.6.0 and higher, if \pkg{bayesplot} (or a package that imports \pkg{bayesplot} such as \pkg{rstanarm} or \pkg{brms}) is loaded, \code{pp_check()} is also available as an alias for \code{check_predictions()}`

bwiernik:
@strengejacke I think we should merge this before CRAN release. Can I go ahead and do that?

strengejacke:
Yes, we can merge this PR.

R/pp_check.R
```diff
@@ -52,18 +49,20 @@
...
-#'   pp_check(model)
+#' model <- lm(mpg ~ disp, data = mtcars)
+#' if (require("see")) {
+#'   check_predictions(model)
 #' }
 #' @export
-pp_check <- function(object, ...) {
-  UseMethod("pp_check")
+check_predictions <- function(object, iterations = 50, check_range = FALSE, re_formula = NULL, ...) {
+  if (isTRUE(insight::model_info(object, verbose = FALSE))) {
```
strengejacke, line 58: I just realized this code line here - what are you expecting from `model_info()` to be returned? I think this if-statement currently is never true.

strengejacke: @bwiernik I think this should be checking for Bayesian methods, right?

PR #362, opened 2021-09-06
Add CI to r2_loo()

bwiernik:
Method pulled from *rstanarm*


strengejacke:
`sweep()` is one of the remaining R mysteries to me ;-)
Should this PR merged before we submit a new version to CRAN? I'm preparing CRAN submission of _performance_ for the next days.

bwiernik:
The computational code is lifted from rstanarm. Here, sweep is used to multiple each column of `wts` by either the observed response values or the posterior draws. This is used to compute row-wise variances.

I think we can merge this before CRAN.

bwiernik:
Realized that *rstantools* has the same function implemented, so I changed our function to just wrap that in the same way as with `r2_bayes()`.

bwiernik:
Okay @strengejacke This now passes all the tests and checks, so I think we can merge before CRAN release

Issue #255, opened 2021-04-03
Information criteria differences and weights 


bwiernik:
It would be really nice if either `compare_performance()` or a new `test_ic()` function could compute the difference between information criteria statistics (e.g., AIC) as well as comparison weights, ala `bbmle::ICtab()`. See [here](https://github.com/bbolker/bbmle/blob/master/R/IC.R).

For example:
```
a <- lm(formula = mpg ~ wt + qsec + am, data = mtcars)
b <- lm(formula = mpg ~ wt + qsec, data = mtcars)
n <- lm(formula = mpg ~ 1, data = mtcars)

bbmle::ICtab(a, b, n, base = TRUE, weights = TRUE)
```

DominiqueMakowski:
We discussed the idea of having a way to display differences rather than raw indices (https://github.com/easystats/performance/issues/28#issuecomment-760889762) either through a different function or an argument, but @strengejacke raised concerns whether this applied (and was meaningful) for all indices. Or should we just apply the difference to a subset of pre-determined metrics?

bwiernik:
I think a `test_ic()` function with an argument to specify the information criterion makes the most sense.

PR #363, opened 2021-09-13
Add IC weights compare_performance function.

bwiernik:
This adds model comparison/stacking weights based on information criteria to the `compare_performance()` function

Closes https://github.com/easystats/performance/issues/255

### DIFF
--- a/R/compare_performance.R
+++ b/R/compare_performance.R
@@ -120,13 +120,13 @@ compare_performance <- function(..., metrics = "all", rank = FALSE, verbose = TR
     dat <- model_performance(.x, metrics = metrics, verbose = FALSE)
     model_name <- gsub("\"", "", .safe_deparse(.y), fixed = TRUE)
     perf_df <- data.frame(Name = model_name, Model = class(.x)[1], dat, stringsAsFactors = FALSE)
-    attributes(perf_df) <- c(attributes(perf_df), attributes(dat)[! names(attributes(dat)) %in% c("names", "row.names", "class")])
+    attributes(perf_df) <- c(attributes(perf_df), attributes(dat)[!names(attributes(dat)) %in% c("names", "row.names", "class")])
     return(perf_df)
   }, objects, object_names, SIMPLIFY = FALSE)
 
   attri <- lapply(m, function(x) {
     attri <- attributes(x)
-    attri[! names(attri) %in% c("names", "row.names", "class")]
+    attri[!names(attri) %in% c("names", "row.names", "class")]
   })
   dfs <- Reduce(function(x, y) merge(x, y, all = TRUE, sort = FALSE), m)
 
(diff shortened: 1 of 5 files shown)


R/compare_performance.R
```diff
@@ -103,11 +119,29 @@ compare_performance <- function(..., metrics = "all", rank = FALSE, verbose = TR
   m <- mapply(function(.x, .y) {
     dat <- model_performance(.x, metrics = metrics, verbose = FALSE)
     model_name <- gsub("\"", "", .safe_deparse(.y), fixed = TRUE)
-    cbind(data.frame(Name = model_name, Model = class(.x)[1], stringsAsFactors = FALSE), dat)
+    perf_df <- data.frame(Name = model_name, Model = class(.x)[1], dat, stringsAsFactors = FALSE)
+    attributes(perf_df) <- c(attributes(perf_df), attributes(dat)[! names(attributes(dat)) %in% c("names", "row.names", "class")])
```
strengejacke: `[! names(attributes(dat))` - here's a whitespace that could be removed.

R/compare_performance.R
```diff
@@ -103,11 +119,29 @@ compare_performance <- function(..., metrics = "all", rank = FALSE, verbose = TR
...
 
+  if (any(c("AIC", "AICc", "BIC", "WAIC") %in% names(dfs))) {
+    dfs$wt_AIC  <- .ic_weight(dfs$AIC)
+    dfs$wt_AICc <- .ic_weight(dfs$AICc)
+    dfs$wt_BIC  <- .ic_weight(dfs$BIC)
+    dfs$wt_WAIC <- .ic_weight(dfs$WAIC)
+  }
+
+  if ("LOOIC" %in% names(dfs)) {
+    lpd_point <- do.call(cbind, lapply(attri, function(x) x$loo$pointwise[,"elpd_loo"]))
```
strengejacke: `[,"elpd_loo"]` - missing whitespace after comma here

strengejacke (changes requested):

strengejacke:
Should we add an argument that returns weighted IC, or should it be returned by default?

R/compare_performance.R
```diff
@@ -103,11 +119,29 @@ compare_performance <- function(..., metrics = "all", rank = FALSE, verbose = TR
...
     model_name <- gsub("\"", "", .safe_deparse(.y), fixed = TRUE)
-    cbind(data.frame(Name = model_name, Model = class(.x)[1], stringsAsFactors = FALSE), dat)
+    perf_df <- data.frame(Name = model_name, Model = class(.x)[1], dat, stringsAsFactors = FALSE)
+    attributes(perf_df) <- c(attributes(perf_df), attributes(dat)[! names(attributes(dat)) %in% c("names", "row.names", "class")])
+    return(perf_df)
   }, objects, object_names, SIMPLIFY = FALSE)
 
+  attri <- lapply(m, function(x) {
+    attri <- attributes(x)
+    attri[! names(attri) %in% c("names", "row.names", "class")]
```
strengejacke: same here - no whitespace after `!`

strengejacke (changes requested):

R/compare_performance.R
bwiernik: I can remove it, but I usually like to keep a space in `! x %in% y` constructions to avoid accidentally removing the `!` on later edits.

bwiernik:
> Should we add an argument that returns weighted IC, or should it be returned by default?

It's not a weighted IC, it's *model weights* based on the IC. For example, what you would use in model averaging/stacking or to determine relative support of each model based on the IC (similar to a Bayes factor). I think its reasonable to just compute these by default when there are information criteria in the metrics.

PR #365, opened 2021-09-15
Add performance_cv function

bwiernik:
This is a bit of a work in progress. It is currently only tested with linear models, but should generally any model that isn't binomial. Cross-validated deviance/loglikelihood isn't implemented yet, and there are some speed ups that could be implemented for normal models when doing loo (that has a closed form solution). 

But otherwise, this should work fairly well.

- [x] print method
- [ ] testing on non-normal models
- [x] speedups for loo on normal models
- [ ] cross-validated deviance


bwiernik:
This is good enough for now. We can add deviance/elpd (log-likelihood) and support for nonlinear models later.

strengejacke:
cool, thanks!

Issue #379, opened 2021-11-25
R^2 measures decomposition?


akarlinsky:
There are several packages out there, like `rego` for Stata that provide a Shapley/Owen decomposition for R^2 and R^2 types (adjusted etc.) goodness-of-fit measures. 

See this paper for example: 
https://projecteuclid.org/journals/electronic-journal-of-statistics/volume-6/issue-none/Axiomatic-arguments-for-decomposing-goodness-of-fit-according-to-Shapley/10.1214/12-EJS710.full

and rego package page:
http://www.marco-sunder.de/stata/rego.html

bwiernik:
Related. https://github.com/easystats/parameters/issues/479

fmg-jluchman:
Acknowledging @bwiernik 's concern about importance metrics and their use- and situation-specific interpretation, if the `performance` package authors are potentially interested Shapley and Owen decomposition, I have a relatively new package I have been working on (under my other alias @jluchman here: https://github.com/jluchman/domir) and would be happy to share/adapt to work within the `easystats` suite of packages.  

Technically, `domir` is a dominance analysis package (as its focus is on dominance designations) but the general dominance statistics it produces are equivalent to Shapley values (and Owen decomposition using the `sets` argument).  This package's goal is to move beyond linear regression (focus of `yhat` and `relaimpo`) to apply Shapley value decomposition and dominance analysis to any statistical or machine learning model's fit to data (and hence could use any/all of the metrics offered by `performance`)

The package I humbly offer for consideration is still very much in development but, if interested, would be happy to collaborate with the authors here to pull/wrap it in and add needed caveats to contextualize the kinds of insights it may provide for the user by way of the insights it adds about the model's results.

fmg-jluchman:
Pleased to hear it!  Look forward to collaborating to bring this in - will start on a fork

DominiqueMakowski:
That sounds great indeed :) Looking forward to that!

From a design standpoint, I don't know what you have in mind for bringing closer together easystats and domir, however since your package looks to be already functional and well documented, I could imagine several possible scenarios:
 
1) *performance* wraps around *domir* and uses it for some of its functions (*domir* becomes a soft dependency of *performance*).
[...]

PR #411, opened 2022-04-03
Add `dominance_analysis` function

jluchman:
Pull request in response to issue #379.

Comments on implementation are discussed in the issue thread.  At a high-level, the function uses `{insight}` to parse and structure model facets for `domir::domin`.  The dominance analysis uses the R2 values from `r2` as applied to all underlying calls to the model.  The print method for `dominance_analysis` has also been updated (which is likely to also be applied to `domir::domin`'s print method) in an attempt to more easily follow the results produced by the function.

Tried to design this function to follow the form/structure of other functions (in particular, `r2`) in `{performance}`.  That said, I have no strong feelings about keeping the structure as is and happy to work to restructure the function (and to take comments/critique on the underlying `domir::domin` function as well) to fit with the norms of other functions in the package.


strengejacke:
Hi Joseph, thanks a lot for this PR!
I did some stylistic changes, so the code style better machtes our conventions. Now I would like to call on @DominiqueMakowski if he can contribute to finding an alias, maybe something like `check_dominance()`, if this would fit? I don't know much about the approach itself, so I cannot comment on the correct implementation of that method, but it looks good to me (we're wrapping the _domir_ package anyway, so there shouldn't be much wrong here).

One thing we could think about it probably revising the print method, so the output looks more "easystats"-like, but that's not urgent nor absolutely necessary...

``` r
library(domir)
library(performance)
model <- glm(vs ~ cyl + carb + mpg, data = datasets::mtcars,
             family = binomial())

r2(model)
#> # R2 for Logistic Regression
#>   Tjur's R2: 0.741
dominance_analysis(model)
#> # Dominance Analysis Results
#> 
#> Model R2 Value: 0.741 
#> 
#> General Dominance Statistics
#>      General Dominance Standardized Ranks
#> cyl              0.366        0.493     1
#> carb             0.178        0.241     3
#> mpg              0.197        0.266     2
#> 
#> Conditional Dominance Statistics
#>      Preds: 1 Preds: 2 Preds: 3
#> cyl     0.654    0.254    0.190
#> carb    0.394    0.066    0.074
#> mpg     0.474    0.085    0.032
#> 
#> Complete Dominance Designations
#>      Dominated by: cyl Dominated by: carb Dominated by: mpg
#> cyl               <NA>              FALSE             FALSE
#> carb              TRUE               <NA>              <NA>
#> mpg               TRUE               <NA>              <NA>
```

<sup>Created on 2022-04-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
Hmm, I actually think this would be better in parameters if possible. People typically report dominance weights in parallel to the regression weights. In that case, we could do `dominance_analysis()` in parallel to `factor_analysis()`

strengejacke:
Yeah, I'm not sure, as I'm, not familiar with that method. So, is it closer to a factor analysis, or closer to a fit-measure like R2?

bwiernik:
It's mostly used as an aid to better interpret the regression coefficients--which is the strongest predictor, etc

bwiernik:
Here is an example image

bwiernik:
This looks great! 

A few requests if possible:

1. Could you add an argument to request all or just general/conditional/complete dominance statistics? Default to all and save it as an attribute. Have an argument in print to control which are printed, defaulting to the attribute.

2. I would be nice to have a method to control the subsets of variables. For example, if I only want to examine dominance for the focal predictors and not the baseline controls. Or if I want to ensure that the higher order terms of a polynomial or interaction are never present without their lower terms. I think maybe take a list, with elements being formulas of elements to add or drop. This could default to NULL to handle each term separately. Then `list(~ a, ~ b)` would mean modulate a and  b (but not other terms in the model), `list(~ a, ~ b + c)` would mean modulate a and the fixed set of b+c, `list(~ a, ~ b, poly(b, 2))` would mean modulate a and b and b^2, but always ensure that b is in the model if b^2 is

bwiernik (changes requested):
See my comment above.

DominiqueMakowski:
> Hmm, I actually think this would be better in parameters if possible.

I agree, as the indices concern the different parameters :)

strengejacke:
I think moving this PR to _parameters_ wouldn't be that difficult, it's mainly copying the main and related test file, and adding the package to SUGGEST in the description file.

fmg-jluchman:
Hi All,

  Understood regarding migrating to `parameters` given the discussion above.  Happy to do so on my end by forking it and making a pull request there once we've worked through requests here.  Please let me know if the team has an alternative preferred approach for migration.

  @strengejacke , happy to adjust the print method taking inspiration now from `model_parameters()`.

  @bwiernik , also happy to add in restrictions for calling specific dominance statistics/designations as well as ways to accommodate sets of predictors and covariates.  I like your suggestion regarding doing so as a list of formulas and will likely implement using that approach.

bwiernik:
Sounds great

bwiernik:
@jluchman Can you move this PR over to parameters, and I can continue to look it over there?

bwiernik:
Closing in favor ofhttps://github.com/easystats/parameters/pull/710

PR #418, opened 2022-04-12
use `convert_data_to_numeric()`

### LINKED ISSUE #397: Remove internal `.factor_to_numeric()` helper?
We have `factor` method for `data_to_numeric` generic in datawizard:
https://easystats.github.io/datawizard/reference/convert_data_to_numeric.html

Can it be used instead of this internal helper? Or this helper has a different logic to it?

https://github.com/easystats/performance/blob/ed1a3984fb0823680956cdb5258cb95e6df06a47/R/helpers.R#L74-L108

IndrajeetPatil:
closes #397

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,10 +61,12 @@ Depends:
 Imports:
     bayestestR (>= 0.11.5),
     insight (>= 0.17.0),
-    datawizard (>= 0.3.0),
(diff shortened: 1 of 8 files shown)


strengejacke:
See https://github.com/easystats/performance/issues/397#issuecomment-1057748021

The default behaviour should be setting `preserve_labels = TRUE`, unless `lowest` is specified.

PR #425, opened 2022-05-24
`fixest` AIC BIC used to hang with default methods

vincentarelbundock:
`model_performance` hangs on some models.

We should use `stats::AIC` and `stats::BIC` directly withn these models


strengejacke:
ok, sounds good! But do you have a reprex where _fixest_ doesn't work with `insight::get_loglikelihood()` (which should result in "hanging")?

vincentarelbundock:
I'm fine either way, but would just ask that the commit be reversed only after we find an adequate solution. As things stand, `model_performance` is unusable with `fixest` models, even if the size is not enormous.

bwiernik:
@strengejacke Can you possibly profile the adjustment function and see which part of it is so slow?

strengejacke:
> the issue isn't the weights, but ensuring that the AIC is comparable to untransformed response

Yes, it's just that we called the jacobian-method (which uses `stats::numDeriv()`) in general, because it handles weights and gives similar results to the `dnorm()` approach by @mattansb. I can try to profile, but I bet it's numDeriv.

bwiernik:
One quick solution would be to skip numDeriv() and use analytic derivatives for common transformations like identity/none, log, exp, and sqrt

strengejacke:
``` r
library(fixest)
library(insight)
set.seed(123)
d <- dplyr::sample_frac(trade, .01)
mod_trade <- feols(log(Euros)~ log(dist_km) | Product + Destination + Origin, data = d)

microbenchmark::microbenchmark(
  adjustment = get_loglikelihood(mod_trade, check_response = TRUE, debug = TRUE),
  jacobian = get_loglikelihood(mod_trade, check_response = TRUE, debug = FALSE),
  times = 10
)
#> Unit: milliseconds
#>        expr       min        lq       mean     median        uq       max neval
#>  adjustment   34.2308   43.9059   57.38631   57.66645   74.4601   75.2174    10
#>    jacobian 1053.2760 1158.6418 1239.35472 1205.24905 1311.4714 1567.5814    10
#>  cld
#>   a 
#>    b
#>

# dlnorm AIC
get_loglikelihood(mod_trade, check_response = TRUE, debug = TRUE)
#> 'log Lik.' -7172.48 (df=2)

# jacobian AIC
get_loglikelihood(mod_trade, check_response = TRUE)
#> 'log Lik.' -6458.599 (df=2)

AIC(mod_trade)
#> [1] 1522.331
```

<sup>Created on 2022-05-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
identity(x) = x
log(x) = 1/x
log(x, base = a) = 1/(x * log(a))
exp(x) = exp(x)
sqrt(x) = .5 / sqrt(x)
1/x = -1/(x^2)

strengejacke:
Seems to work fine!

``` r
library(fixest)
library(insight)
set.seed(123)
d <- dplyr::sample_frac(trade, .01)
m <- feols(Euros~ log(dist_km) | Product + Destination + Origin, data = d)
get_loglikelihood(m, check_response = TRUE)
#> 'log Lik.' -7556.083 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "analytic")
#> 'log Lik.' -7556.083 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "dlnorm")
#> 'log Lik.' -7556.083 (df=2)

m <- lm(exp(hp) ~ cyl, data = mtcars)
get_loglikelihood(m, check_response = TRUE)
#> 'log Lik.' -6014.888 (df=3)
get_loglikelihood(m, check_response = TRUE, debug = "analytic")
#> 'log Lik.' -6014.888 (df=3)
get_loglikelihood(m, check_response = TRUE, debug = "dlnorm")
#> 'log Lik.' -6014.888 (df=3)

m <- feols(log(Euros)~ log(dist_km) | Product + Destination + Origin, data = d)
get_loglikelihood(m, check_response = TRUE)
#> 'log Lik.' -6458.599 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "analytic")
#> 'log Lik.' -6458.599 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "dlnorm")
#> 'log Lik.' -7172.48 (df=2)

m <- feols(sqrt(Euros)~ log(dist_km) | Product + Destination + Origin, data = d)
get_loglikelihood(m, check_response = TRUE)
#> 'log Lik.' -6767.562 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "analytic")
#> 'log Lik.' -6767.562 (df=2)
get_loglikelihood(m, check_response = TRUE, debug = "dlnorm")
#> 'log Lik.' -13403.38 (df=2)
```

<sup>Created on 2022-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

For any none-supported transformation, jacobian-derivative function is called.

@vincentarelbundock Can you revert this PR, since the function now is much faster again?

vincentarelbundock:
Fantastic! 

When you say "revert the PR", do you just want me to call this?

```bash
git reset --hard HEAD~1
git push upstream master --force
```

bwiernik:
Just do a new revert commit since it's already merged to master. 

If you use the GitHub desktop app, right click on the commit and choose revert. Or else https://www.freecodecamp.org/news/git-revert-commit-how-to-undo-the-last-commit/

bwiernik:
sqrt() looks like it has an extra log() in the derivative, and what's the case of NULL weights doing?

vincentarelbundock:
thanks. revert is done.

strengejacke:
Ok, but see https://github.com/easystats/performance/pull/425#issuecomment-1136515744, all results are identical. So it seems the implementation in the ll-adjustment that uses "numericDeriv" is wrong?

Can you check [`insight::.ll_analytic_adjustment ()`](https://github.com/easystats/insight/blob/a694a27918ab440e5fdad8b167d28d2f2f7f1dac/R/get_loglikelihood.R#L423)? I guess, all `log()` need to be removed (except for log-transformation...).

bwiernik:
My bad. Everything is fine. I was misremembering exactly how we used the output for these.

vincentarelbundock:
Sorry, I was going to fix the issue on `modelsummary` tracker and tried again, but the problem seems to be there still. Do I need to do anything to merge the analytic fix?

strengejacke:
_insight_

vincentarelbundock:
Right, thanks a lot.

Unfortunately, `performance(mod_trade, metric = "bic")` still takes forever. Is this the same problem?

vincentarelbundock:
Sorry for bugging you. I figured it out, copied over the `ll_analytic_adjust()` function from `insight`, and now it's super fast.

BTW, is this adjustment documented anywhere?

strengejacke:
Thanks! I'm, not sure, there was a reason why we didn't use `BIC(get_loglikelihood())` in all places, so indeed the fix in _insight_ should also be copied over to here.

It's not very offensively documented, but you find a small sentence here:
https://easystats.github.io/performance/reference/performance_aicc.html

and here:
https://easystats.github.io/performance/reference/test_performance.html#tests-description

Maybe we could promote this a bit stronger?

strengejacke:
@bwiernik @vincentarelbundock Can you please check following code?

In [this file](https://github.com/easystats/performance/blob/master/R/helpers.R), we have `.get_BIC()`. At the bottom, we use

```r
stats::BIC(insight::get_loglikelihood(x, check_response = TRUE, REML = REML, verbose = FALSE))
```

i.e. we account for transformed response variables. Yet, the last line calls the internal adjustment again:

```r
  .adjust_ic_jacobian(x, out)
```

I think the last line with the call to `.adjust_ic_jacobian()` can/should be removed, right?

```r
.get_BIC <- function(x, estimator = "ML") {
  # check ML estimator
  if (missing(estimator) && inherits(x, "lmerMod")) {
    estimator <- "REML"
  }
  REML <- identical(estimator, "REML")

  if (inherits(x, c("vgam", "vglm"))) {
    insight::check_if_installed("VGAM")
    out <- VGAM::BIC(x)
  } else if (inherits(x, "bayesx")) {
    out <- stats::BIC(x)[["BIC"]]

  } else {
    out <- tryCatch(
      stats::BIC(insight::get_loglikelihood(x, check_response = TRUE, REML = REML, verbose = FALSE)),
      error = function(e) NULL
    )
  }
  .adjust_ic_jacobian(x, out)
}
```

bwiernik:
Yes, that should be removed. I am guessing that's a holdover from your first implementation

vincentarelbundock:
Not sure. In the previous implementation there were *both* calls to `.adjust_ic_jacobian` and to `insight::get_loglikelihood`, and my recollection is that `get_loglikelihood` was already doing the correction, only numerically instead of analytically (which was slow). So were we correcting twice?

Here's the old blob:

https://github.com/easystats/performance/blob/445af864b8d09dcdd73ff5b7399badc464add565/R/helpers.R#L15

strengejacke:
Not all functions have a logLik-method, and e.g. only return the AIC. In such cases, we need the ic-adjustment.

vincentarelbundock:
Woke up this morning with the intention to finish this, but in classic Daniel fashion, he appears to have gotten to it while I was sleeping.

Thanks!

PR #426, opened 2022-05-25
norm for numeric

mattansb:
What do you think?

``` r
library(performance)
x <- rt(100, 2)

(pnorm <- check_normality(x))
#> Warning: Non-normality of raw detected (p = 0.003).
  
plot(pnorm)
```

``` r
plot(pnorm, type = "qq")
```

<sup>Created on 2022-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


strengejacke:
Nice idea! 👍

IndrajeetPatil:
I don't get it. What's the issue here?

S3 method seems to be getting registered as expected.

PR #442, opened 2022-06-29
Norm for numeric

strengejacke:


R/check_normality.R
```diff
@@ -68,6 +68,32 @@ check_normality.default <- function(x, ...) {
...
+  attr(p.val, "data") <- x
+  attr(p.val, "object_name") <- insight::safe_deparse(substitute(x))
+  attr(p.val, "effects") <- "fixed"
+  class(p.val) <- unique(c("check_normality", "see_check_normality", "check_normality_numeric", class(p.val)))
+
+  p.val
+}
+
+#' @export
+#' @importFrom stats residuals
```
bwiernik: If we wanted to not import from stats, we can use `@s3method` instead

bwiernik: Compare to how we do pp_check

strengejacke: like this?

bwiernik (approved):

PR #445, opened 2022-07-02
ci=NULL speeds up computation

vincentarelbundock:
When we don't need confidence intervals, it is much faster to skip their computation.


bwiernik (approved):

IndrajeetPatil:
Thanks!

Issue #449, opened 2022-07-14
Are there plans to support factor (coded) outcomes for logistic regression?


krassowski:
Let's take the example from readme:

```R
model <- glm(am ~ wt + cyl, data = mtcars, family = binomial)
check_model(model)
```

This one works fine. Let's create a new column in mtcars which will code am as "automatic" or "manual" for ease of interpretation:

```R
mtcars$transmission_type = relevel(
    as.factor(ifelse(mtcars$am, 'automatic', 'manual')),
    ref='manual'
)
factor_model <- glm(transmission_type ~ wt + cyl, data = mtcars, family = binomial)
check_model(factor_model)
```

This results in:

```
Error in `$<-.data.frame`(`*tmp*`, "group", value = "yes"): replacement has 1 row, data has 0
Traceback:

1. check_model(model)
2. check_model.default(model)
3. suppressWarnings(.check_assumptions_glm(x, minfo))
4. withCallingHandlers(expr, warning = function(w) if (inherits(w, 
 .     classes)) tryInvokeRestart("muffleWarning"))
5. .check_assumptions_glm(x, minfo)
6. binned_residuals(model)
7. `$<-`(`*tmp*`, "group", value = "yes")
8. `$<-.data.frame`(`*tmp*`, "group", value = "yes")
9. stop(sprintf(ngettext(N, "replacement has %d row, data has %d", 
 .     "replacement has %d rows, data has %d"), N, nrows), domain = NA)
```

This can also be reproduced by running `binned_residuals(factor_model)`.

Similarly, `r2(factor_model)` does not work:

```
Warning message in .set_back_labels(as.numeric(as.character(x)), x):
“NAs introduced by coercion”
Warning message in min(x, na.rm = TRUE):
“no non-missing arguments to min; returning Inf”
$R2_Tjur = Tjur's R2: NaN
```

bwiernik:
Both of those work fine for me. Install the latest versions of the easystats packages and try again:

```r
install.packages("easystats", repos = "https://easystats.r-universe.dev")
easystats::install_latest()
```

krassowski:
Thank you for testing. Unfortunately, it still does not work for me. I have run the suggested command which upgraded the following packages:

| Package     |   Latest | Installed |
| --------------|--------|------------ |
| insight     | 0.18.0.1 |    0.18.0 |
| datawizard  |  0.4.1.8 |   0.4.1.7 |
| performance |  0.9.1.1 |     0.9.1 |
[...]

PR #451, opened 2022-07-14
Add test for factor response

krassowski:
Should help to prevent regressions like one discussed in #449.


IndrajeetPatil:
Thanks!

PR #459, opened 2022-08-02
check_convergence.glmmTMB

DominiqueMakowski:
As a matter of fact I would have expected `is_converged()` to be in insight. I'd move the function there, and then let `check_convergence()` be a simple wrapper around it. 

For instance, `is_converged()` could return a boolean (makes it easy to program with), while `check_convergence()` could, in the future, provide some more diagnostic details about what failed, what to do etc. (similar to `glmmTMB::diagnose()`)


DominiqueMakowski:
note that it seems like accessing convergence success is not super straightforward in glmmTMB (https://github.com/glmmTMB/glmmTMB/issues/275)

strengejacke:
> As a matter of fact I would have expected is_converged() to be in insight. I'd move the function there, and then let check_convergence() be a simple wrapper around it.

Done.

PR #460, opened 2022-08-02
Use new pkgdown template

etiennebacher:
https://github.com/easystats/easystats/issues/186


IndrajeetPatil:
pkgdown GHA is also failing in the default branch.

I think this is related to the recent breaking changes in _datawizard_. One of the dependencies must be out of date.,

PR #492, opened 2022-10-17
fix: rm additional backticks messing with styling

mcanouil:
This PR removes a typo where an extra backtick was added which messed up the styling in documentation.


IndrajeetPatil (approved):
Thanks!

Issue #434, opened 2022-06-14
May "conditional ICC" not being *conditional*?


iago-pssjd:
I wonder if calling conditional to the non-adjusted value reported by `icc` is suitable. It may produce some confussion as you may look at the example below.

Looking at *Nakagawa et al. 2017*, they do not mention anywhere "conditional ICC", but just an ICC adjusted and another ICC not including any subscript to the symbol, as they do for the adjusted or for the marginal and conditional R^2 (which I imagine is the reason you talk about conditional ICC).

Further, in the book *Rabe-Hesketh and Skrondal (2012)*, they talk about a conditional or residual intraclass correlation, but it agrees (!!!) with the "adjusted ICC". Indeed, looking at the examples in section 3.5, specifically at the last paragraph before section 3.6 (page 137), they say:

> For the intraclass correlation, we see that the unconditional intraclass correlation for the null model without covariates is estimated as 0.487. This reduces to a conditional or residual intraclass correlation of 0.459 when level-2 covariates are added and to 0.455 when all remaining covariates are added.

And if we download the data from http://www.stata-press.com/data/mlmus4.html and replicate the analysis (that agree with tables in pages 132 [or 134] and 137):
```r
library(haven)
smoking <- read_dta("mlmus4/smoking.dta")
library(lme4)
null <- lmer(birwt ~ (1|momid), data = smoking)
all <- lmer(birwt ~ (1|momid) + smoke + male + mage + hsgrad + somecoll + collgrad + married + black + kessner2 + kessner3 + novisit + pretri2 + pretri3, data = smoking)
fit <- lmer(birwt ~ (1|momid) + hsgrad + somecoll + collgrad + married + black, data = smoking)
library(parameters)
library(performance)
model_parameters(fit)
Package 'merDeriv' needs to be installed to compute confidence intervals for random effect parameters.
# Fixed Effects 

Parameter   | Coefficient |    SE |             95% CI | t(8596) |      p
-------------------------------------------------------------------------
(Intercept) |     3216.48 | 25.84 | [3165.82, 3267.14] |  124.45 | < .001
hsgrad      |      131.44 | 24.93 | [  82.57,  180.31] |    5.27 | < .001
somecoll    |      180.69 | 26.52 | [ 128.70,  232.68] |    6.81 | < .001
collgrad    |      232.90 | 25.61 | [ 182.70,  283.09] |    9.10 | < .001
married     |      114.76 | 25.48 | [  64.82,  164.71] |    4.50 | < .001
[...]

PR #495, opened 2022-10-18
[Docs] minor ICC docs improvement

DominiqueMakowski:
I want to add a paragraph on the difference between ICC and R2, because that's a question that keeps coming back. But unfortunately, I can't find anything concrete...

### DIFF
--- a/R/check_outliers.R
+++ b/R/check_outliers.R
@@ -90,16 +90,15 @@
 #'  (`"zscore"`) or, as it is here the case (`"zscore_robust"`) by
 #'  default (Iglewicz, 1993), in terms of Median Absolute Deviation (MAD) from
 #'  the median (which are robust measures of dispersion and centrality). The
-#'  default threshold to classify outliers is 1.959 (`threshold =
-#'  list("zscore" = 1.959)`), corresponding to the 2.5\% (`qnorm(0.975)`)
-#'  most extreme observations (assuming the data is normally distributed).
-#'  Importantly, the Z-score method is univariate: it is computed column by
-#'  column. If a dataframe is passed, the Z-score is calculated for each
-#'  variable separately, and the maximum (absolute) Z-score is kept for each
-#'  observations. Thus, all observations that are extreme on at least one
-#'  variable might be detected as outliers. Thus, this method is not suited for
-#'  high dimensional data (with many columns), returning too liberal results
-#'  (detecting many outliers).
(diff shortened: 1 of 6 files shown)


R/check_outliers.R
```diff
@@ -90,16 +90,15 @@
...
-#'  list("zscore" = 1.959)`), corresponding to the 2.5\% (`qnorm(0.975)`)
-#'  most extreme observations (assuming the data is normally distributed).
-#'  Importantly, the Z-score method is univariate: it is computed column by
-#'  column. If a dataframe is passed, the Z-score is calculated for each
-#'  variable separately, and the maximum (absolute) Z-score is kept for each
-#'  observations. Thus, all observations that are extreme on at least one
-#'  variable might be detected as outliers. Thus, this method is not suited for
-#'  high dimensional data (with many columns), returning too liberal results
-#'  (detecting many outliers).
+#'  default threshold to classify outliers is 1.959 (`threshold = list("zscore" = 1.959)`),
```
DominiqueMakowski, line 93: this change just puts the code expression on the same line so that it renders correctly

R/icc.R
```diff
@@ -52,10 +52,21 @@
...
+#'  models with more complex random effects structures. The ICC can help determine
+#'  whether a mixed model is even necessary: an ICC of zero means the
+#'  observations within clusters are no more similar than observations from
+#'  different clusters, and setting it as a random factor might not be necessary.
+#'  }
+#'  \subsection{Difference with R2}{
+#'  The coefficient of determination R2 (that can be computed with `r2()`)
+#'  quantifies the proportion of variance explained by a statistical model, but
+#'  its definition in mixed model is complex (hence, different methods to compute
+#'  a proxy exist). ICC is related to R2 because they are both ratios of
```
DominiqueMakowski, line 68: Here... even in the nakagawa paper it only says "they are related" lol

strengejacke:
See https://github.com/easystats/performance/issues/434
We should probably not use the conditional ICC (which also is better named "unadjusted" ICC).

I would propose: R2 is the proportion of the explained variance, while the ICC is the proportion of explained variance that can be accounted to the random effects.

The difference between the conditional and marginal R2 is just the difference between the fixed and random effects variances (as defined [here](https://easystats.github.io/insight/reference/get_variance.html)).

strengejacke (approved):

PR #508, opened 2022-10-30
Implement R2 CI

### LINKED ISSUE #504: CIs for R2?
In the docs for `r2()`, I saw we mention the `ci` argument, where CIs can be based on bootstrap or we calculate analytical CIs for R2 (https://easystats.github.io/performance/reference/r2.html).

But do we actually support this anywhere?

@easystats/core-team

strengejacke:
Fixes #504


strengejacke:
@bwiernik can we use `.r2_ci()` also for the different pseudo-r2, like mcfadden, nagelkerke etc.?

PR #511, opened 2022-10-31
Add tests for check_outliers

rempsyc:
First attempt at writing tests for easystats so welcome any feedback. I removed the `if (requiet("testthat") && requiet("performance"))` at the beginning since we discussed it wasn't necessary earlier since performance will already be installed and testthat as well.

### DIFF
--- a/R/check_outliers.R
+++ b/R/check_outliers.R
@@ -1201,9 +1201,8 @@ check_outliers.gls <- function(x,
                                ID = NULL,
                                ...) {
   if (!missing(ID)) {
-    warning(
-      paste0("ID argument not supported for objects of class `", class(x)[1], "`."),
-      call. = FALSE
+    insight::format_warning(
+      paste0("ID argument not supported for objects of class `", class(x)[1], "`.")
     )
   }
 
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
Thansk, @rempsyc! 

Can you please have a look at the build failures? They are not to be expected.

strengejacke (approved):

strengejacke:
Looks like one issue has to be fixed.

IndrajeetPatil:
The error is the following:

```r
 ----------- FAILURE REPORT -------------- 
 --- failure: the condition has length > 1 ---
 --- srcref --- 
 at D:/a/performance/performance/check/performance.Rcheck/00_pkg_src/performance/R/check_outliers.R#1216: 
```

IndrajeetPatil:
```r
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test-check_outliers.R:295'): nlme multiple methods ────────────────
  `check_outliers(model, method = c("zscore_robust", "iqr"))` did not throw the expected error.
  
  [ FAIL 1 | WARN 10 | SKIP 3 | PASS 309 ]
```

strengejacke:
I think we can remove that test, because this error should actually never occur.

rempsyc:
Thanks! The failing test is successful when run locally though that's weird.

Note that while doing the tests, I realized that some tests that should be working were not working, so I have commented them out for now. I did not want to withhold all the other successful tests waiting for me to debug a few specific cases. However, I will come back to them later and attempt to fix them as well.

PR #497, opened 2022-10-20
Fixes #496 (`check_outliers` throws error when passing multiple element to threshold list)

rempsyc:
Fixes #496 (`check_outliers` throws error when passing multiple element to threshold list)

---

``` r
# Load library
devtools::load_all(".../forks/performance")

packageVersion("performance")
#> [1] '0.10.0.4'

check_outliers(mtcars, method = c(
  "zscore", "mahalanobis"), threshold = list(
    'zscore' = 2.5, 'mahalanobis' = 15))
#> 2 outliers detected: cases 9, 31.
#> - Based on the following methods and thresholds: zscore (2.5),
#>   mahalanobis (15).
#> - For variables: mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb.
#> Note: Outliers were classified as such by at least half of the selected methods. 
#> 
#> -----------------------------------------------------------------------------
#> The following observations were considered outliers for two or more variables 
#> by at least one of the selected methods: 
#> 
#>   Row n_Zscore  n_Mahalanobis
#> 1  31        2 (Multivariate)
#> 2   9        1 (Multivariate)
#> 3  27        0 (Multivariate)
#> 4  29        0 (Multivariate)
```

<sup>Created on 2022-10-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #512, opened 2022-11-01
Add tests, fix bug with model-based outliers and multiple thresholds

rempsyc:
In this PR, I've fixed some (commented out) tests that were failing. One of the reasons they were failing was that I needed to also apply the fix from #497 to model-based outliers (to support user-specified thresholds). Once that was done, the tests were working.

### DIFF
--- a/R/check_outliers.R
+++ b/R/check_outliers.R
@@ -375,7 +375,7 @@ check_outliers.default <- function(x,
     thresholds <- .check_outliers_thresholds(data)
   } else if (is.list(threshold)) {
     thresholds <- .check_outliers_thresholds(data)
-    thresholds[[names(threshold)]] <- threshold[[names(threshold)]]
+    thresholds[names(threshold)] <- threshold[names(threshold)]
   } else {
     insight::format_error(
       "The `threshold` argument must be NULL (for default values) or a list containing threshold values for desired methods (e.g., `list('mahalanobis' = 7)`)."
--- a/tests/testthat/test-check_outliers.R
+++ b/tests/testthat/test-check_outliers.R
@@ -257,6 +257,7 @@ test_that("cook multiple methods which", {
 
 if (requiet("rstanarm")) {
   test_that("pareto which", {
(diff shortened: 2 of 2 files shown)


IndrajeetPatil:
```r
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-check_outliers.R:261'): pareto which ─────────────────────────
which(check_outliers(model, method = "pareto", threshold = list(pareto = 0.5))) (`actual`) not equal to 17 (`expected`).

  `actual`: 9 17
`expected`:   17

[ FAIL 1 | WARN 11 | SKIP 3 | PASS 311 ]
```

rempsyc:
Ok, I've corrected that test output. Let's see what happens now.

rempsyc:
Seems like it's working... but weird thing is, locally for that test (directly in console, with `devtools::test()`, or with `devtools::check(args = c('--as-cran','--run-donttest'))`), I need to expect `17` only, not `c(9, 17)`, else it fails. On GHA it's the opposite. Is it common to have consistent inconsistency between local and GHA like that?

I don't think the pareto method is supposed to have an element of randomness to it that could explain this discrepancy...

strengejacke:
Thanks!

strengejacke:
> Seems like it's working... but weird thing is, locally for that test (directly in console, with devtools::test(), or with devtools::check(args = c('--as-cran','--run-donttest'))), I need to expect 17 only, not c(9, 17), else it fails. On GHA it's the opposite. Is it common to have consistent inconsistency between local and GHA like that?

Often, it's an issue of different setups. It's not always clear why results differ, and in particular with Stan, I feel like often failures are not due to errors in my packages, but to the difficult installation setup in the background.

strengejacke:
Locally, it's also only case 17 for me. I added a seed, maybe that might change the test results?

strengejacke:
setting seeds resolves this issue. We should pay attention to always using seed for models with sampling.

PR #514, opened 2022-11-17
AIC and BIC without loglikelihood

vincentarelbundock:
In some models, `get_loglikelihood()` does not work, but `stats::AIC` and `stats::BIC` can still return useable results. For example, fixest:

```r
library(fixest)
mod <- feols(mpg ~ hp, data = mtcars)
stats::AIC(mod)
stats::BIC(mod)
```


strengejacke:
Thanks!

PR #521, opened 2022-12-09
Add package description file

etiennebacher:
https://github.com/easystats/easystats/issues/314


R/performance-package.R
```diff
@@ -22,7 +22,6 @@
 #' References: Lüdecke et al. (2021) \doi{10.21105/joss.03139}
 #'
 #' @docType package
-#' @aliases performance performance-package
 #' @name performance-package
 #' @keywords internal
 "_PACKAGE"
```
IndrajeetPatil:
```suggestion
"_PACKAGE"

```

R/performance-package.R
```diff
@@ -0,0 +1,27 @@
+#' \code{performance-package}
```
IndrajeetPatil, line 1: Does this create problems?

```suggestion
#' \code{performance}
```

R/performance-package.R
etiennebacher: I already added a new line locally but apparently git doesn't show it. I also can't commit your suggestion so I suppose it's not feasible?

R/performance-package.R
etiennebacher: yes it creates a warning in R CMD check:
> Rd files with duplicated alias 'performance':
>     'model_performance.Rd' 'performance-package.Rd'

IndrajeetPatil (approved):
Thanks!

PR #527, opened 2023-01-04
Fix usage of `detectCores()` to get number of cores

### LINKED ISSUE #524: Avoid using `detectCores()` to get number of cores
See: https://www.jottr.org/2022/12/05/avoid-detectcores/

https://github.com/easystats/performance/blob/c3b12bcbd6af76fa09a28589d27f5cf83020344f/R/check_outliers.R#L1566-L1570

strengejacke:
Fixes #524


R/check_outliers.R
```diff
@@ -706,6 +706,7 @@ check_outliers.data.frame <- function(x,
     thresholds[names(threshold)] <- threshold[names(threshold)]
   } else if (is.numeric(threshold)) {
     thresholds <- .check_outliers_thresholds(x)
+    ## FIXME: @rempsyc is this working as intended? 
     thresholds <- lapply(thresholds, function(x) threshold)
```
strengejacke, line 710: Not part of this PR, but @rempsyc is this working as intended?

rempsyc: @strengejacke thanks, I will have a look at it.

rempsyc: I don't know what line 710 does exactly, but providing a numeric threshold seems to work as expected:

``` r
library(performance)
#> Warning: package 'performance' was built under R version 4.2.2

x <- mtcars

# } else if (is.list(threshold)) {
y <- check_outliers(x, threshold = list(mahalanobis = 15))
y
#> 4 outliers detected: cases 9, 27, 29, 31.
#> - Based on the following method and threshold: mahalanobis (15).

# } else if (is.numeric(threshold)) {
z <- check_outliers(x, threshold =  15)
z
#> 4 outliers detected: cases 9, 27, 29, 31.
#> - Based on the following method and threshold: mahalanobis (15).

all.equal(y, z)
#> [1] TRUE
```

<sup>Created on 2023-01-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Was your concern that it was throwing an error in some cases or just that you didn't know if that line was necessary?

rempsyc: Ok, so it's confusing because `threshold` is a parameter, not a function, but essentially, line 710 (`lapply(thresholds, function(x) threshold)`) replaces all threshold values with the numeric value provided (with e.g., `threshold = 15`), resulting in:

``` r
lapply(thresholds, function(x) threshold)
$zscore
[1] 15

$zscore_robust
[1] 15

$iqr
[1] 15

$ci
[1] 15

$hdi
[1] 15

$eti
[1] 15

$bci
[1] 15

$cook
[1] 15

$pareto
[1] 15

$mahalanobis
[1] 15

$mahalanobis_robust
[1] 15

$mcd
[1] 15

$ics
[1] 15

$optics
[1] 15

$lof
[1] 15
```

Which is the correct way, or rather, the way that it has been done in `check_outliers` so far.

PR #528, opened 2023-01-04
issue with item_difficulty()-function

### LINKED ISSUE #517: issue with item_difficulty()-function
In the scientific literature (f.i. p. 76 in Kelava & Moosbrugger, 2012; p. 219 in Bortz & Döring, 2006) it is recommended to calculate the difficulty index (P(Item)) as the quotient of the **actually achieved sumscore** (sum of the actually achieved score of each person who answered the item) and the **maximum achievable score** (product of the maximum possible score * total number of people who answered the item). 
In the calculation of the item_difficulty() function, if I have understood correctly, the max(.x) is used, which corresponds to the **actual maximum** value of the data (item). 
However, when comparing this with a common definition of the item difficulty (see above), I noticed that it is **not** the **actual maximum value** of the data (item) that should be included, but the **possible maximum** value.

To take this into account I have a suggestion: I am sure there a better ways in coding but my first suggestion would be:

```go
item_difficulty <- function(x, max_x = NULL) {
  
  if (is.null(max_x)) {
    d <- sapply(x, function(.x) {
      .x <- .x[!is.na(.x)]
      round(sum(.x) / (max(.x) * length(.x)), 2)
    })
    # ideal item item_difficulty
    fun.diff.ideal <- function(.x) {
      p <- 1 / max(.x, na.rm = TRUE)
      round(p + (1 - p) / 2, 2)
    }
  } else {
    d <- sapply(x, function(.x) {
      .x <- .x[!is.na(.x)]
      round(sum(.x) / (max_x * length(.x)), 2)
    })
    # ideal item item_difficulty
    fun.diff.ideal <- function(.x) {
      p <- 1 / max_x
      round(p + (1 - p) / 2, 2)
    }
  }

  
  

  
  di <- apply(x, 2, fun.diff.ideal)
  
  structure(
    class = c("item_difficulty", "data.frame"),
    data.frame(
      item = colnames(x),
      difficulty = d,
      ideal = di,
      stringsAsFactors = FALSE
    )
  )
}



# methods --------------------------------------

#' @export
print.item_difficulty <- function(x, ...) {
  spaces <- max(nchar(x$item))
  
  insight::print_color("# Item Difficulty\n\n", "blue")
  insight::print_color(sprintf("  %*s  ideal\n", spaces + 10, "difficulty"), "red")
  
  for (i in seq_along(x$item)) {
    cat(sprintf("  %*s      %.2f   %.2f\n", spaces, x$item[i], x$difficulty[i], x$ideal[i]))
  }
  invisible(x)
}
```

In my suggestion the **possible maximum** must be specified with the `max_x`-term. The default is `max_x= NULL` so that your function could be used exactly like before. But who ever wants/needs to specify the possible maximum can do so by defining it. 
I can't think of a way to calculate the **possible maximum** of an item automatically from the item (x), as this information may not be included most data sets.

A problem I see so far with my suggested code, would arise if a data frame including **multiple** items with **different possible maximum** values is used with the function. So far only one possible maximum can be specified with the `max_x`-term which will then be applied to all of the items. 

Sources:
Bortz, J., & Döring, N. (2006). Quantitative Methoden der Datenerhebung. In J. Bortz & N. Döring, Forschungsmethoden und Evaluation (S. 137–293). Springer Berlin Heidelberg. https://doi.org/10.1007/978-3-540-33306-7_4

Kelava, A., & Moosbrugger, H. (2012). Deskriptivstatistische Evaluation von Items (Itemanalyse) und Testwertverteilungen. In H. Moosbrugger & A. Kelava (Hrsg.), Testtheorie und Fragebogenkonstruktion (S. 75–102). Springer Berlin Heidelberg. https://doi.org/10.1007/978-3-642-20072-4_4

strengejacke:
Fixes #517

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,10 @@
 * `test_wald()` now warns the user about inappropriate F test and calls
   `test_likelihoodratio()` for binomial models.
 
+* Fixed issue in `item_difficulty()` with detecting the maximum values of an
+  item set. Furthermore, `item_difficulty()` gets a `maximum_value` argument
+  in case no item contains the maximum value due to missings.
+
 # performance 0.10.0
 
 ## Breaking Change
--- a/R/check_itemscale.R
+++ b/R/check_itemscale.R
@@ -86,10 +86,10 @@ check_itemscale <- function(x) {
 
     s_out <- data.frame(
       Item = columns,
-      Missings = sapply(items, function(i) sum(is.na(i)) / nrow(items)),
-      Mean = sapply(items, mean, na.rm = TRUE),
-      SD = sapply(items, stats::sd, na.rm = TRUE),
-      Skewness = sapply(items, function(i) as.numeric(datawizard::skewness(i))),
+      Missings = vapply(items, function(i) sum(is.na(i)) / nrow(items), numeric(1)),
+      Mean = vapply(items, mean, numeric(1), na.rm = TRUE),
+      SD = vapply(items, stats::sd, numeric(1), na.rm = TRUE),
+      Skewness = vapply(items, function(i) as.numeric(datawizard::skewness(i)), numeric(1)),
       "Difficulty" = item_difficulty(items)$difficulty,
       "Discrimination" = .item_discr,
       "alpha if deleted" = .item_alpha,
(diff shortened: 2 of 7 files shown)


strengejacke:
Needs tests

PR #535, opened 2023-01-12
`check_outliers`: Restored "lof" documentation

### LINKED ISSUE #534: `check_outliers`: Restore "lof" documentation
In commit d3c7add04ae02225db6141cd52f7c1ffbeb1cc6f , we accidentally removed documentation for LOF in addition to isolation forests, but LOF seems to work without problem:

``` r
library(performance)
#> Warning: package 'performance' was built under R version 4.2.2

check_outliers(mtcars, method = "lof", threshold = 0.01)
#> 3 outliers detected: cases 4, 15, 31.
#> - Based on the following method and threshold: lof (0.01).
```

However, we should probably add better error messages when providing thresholds smaller than 0 or bigger than 1:

``` r
check_outliers(mtcars, method = "lof", threshold = -0.01)
#> Warning in stats::qnorm(1 - threshold): NaNs produced
#> OK: No outliers detected.
#> - Based on the following method and threshold: lof (-0.01).
#> - For variables: mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb

check_outliers(mtcars, method = "lof", threshold = 1.01)
#> Warning in stats::qnorm(1 - threshold): NaNs produced
#> OK: No outliers detected.
#> - Based on the following method and threshold: lof (1.01).
#> - For variables: mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

rempsyc:
closes #534 
---
readd lof docs, roxygen2 to rmarkdown syntax, better lof error messages + added tests, better rounding for print method


R/check_outliers.R
```diff
@@ -621,13 +618,10 @@ print.check_outliers <- function(x, ...) {
...
-      (length(method) > 1 || all(method %in% method.univariate))) {
-      cat(long_dash,
-        "The following observations were considered outliers ",
-        "for two or more variables \n",
-        "by at least one of the selected methods: \n\n",
-        sep = ""
-      )
+          (length(method) > 1 || all(method %in% method.univariate))) {
+      cat(long_dash, "The following observations were considered outliers ",
+          "for two or more variables \n", "by at least one of the selected methods: \n\n",
```
strengejacke: We could use `insight::format_alert()` here, which takes care of line breaks taking the terminal width into account.

strengejacke (approved):

DominiqueMakowski:
Maybe we could incorporate some methods mentioned in https://twitter.com/olivier_klein/status/1620874451637403649?s=20&t=xolHxOLDvXVYWOcxjTtSDg, to detect suspicious patterns?

R/check_outliers.R
rempsyc: Ok I just tried it. Right now, the colour of the text (white, because I'm using a dark RStudio theme) matches the colour of the printed data frame. But when using `format_alert`, it changes the colour to blue. I would like it to be white though (or whatever default console colour is), since it's easier to read with lots of text as in our case (blue makes it look like a warning but it's literally the output print method so that's weird). How can I change the colour using `format_alert`? I've tried `format_alert("This is a message.", color = "white")` but it's not working.

strengejacke: Strange, this functions just calls `message()`, so no specific colour should be used except what your ide uses as default colour.

rempsyc: My `message()`s print in blue, that's why I used `cat` originally. So my messages print the same colour as my warnings. Is that normal?

rempsyc: I just tried all the RStudio light themes. All of themes always print warnings and messages with the same colour (so same as with dark themes).

Also, it seems to me that since this is a print method, `suppressMessages()` should not suppress the output. So I don't think using a message is the solution here.

rempsyc: @strengejacke OK, `format_message()` was what we were looking for, it does print correctly ;D, I've corrected the PR

rempsyc:
I think this is ready to merge now.

PR #540, opened 2023-01-16
Suppress messages in tests

### LINKED ISSUE #525: Suppress irrelevant messages in tests
Although there are no more warnings, there are still a number of messages that need to be suppressed.

Follow-up on https://github.com/easystats/easystats/issues/338

Current output: https://github.com/easystats/performance/actions/runs/3796735058/jobs/6457089829

```r
boundary (singular) fit: see help('isSingular')
boundary (singular) fit: see help('isSingular')
Some of the nested models seem to be identical
boundary (singular) fit: see help('isSingular')
refitting model(s) with ML (instead of REML)
```

etiennebacher:
Close #525

### DIFF
--- a/R/check_homogeneity.R
+++ b/R/check_homogeneity.R
@@ -58,7 +58,7 @@ check_homogeneity.default <- function(x, method = c("bartlett", "fligner", "leve
   if (method == "auto") {
     check <- tryCatch(
       {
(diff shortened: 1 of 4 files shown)


strengejacke:
Thanks, looks good!

PR #544, opened 2023-01-18
`check_outliers` paper

rempsyc:
Draft of the `check_outliers` paper (re: https://github.com/orgs/easystats/teams/core-team/discussions/5).

Looking forward to any and all feedback and contributions @easystats/core-team!

PS: reminder to use [skip ci] in your pushes (don't keep forgetting like me!!).

PR #567, opened 2023-03-31
Test cleanup part 2

etiennebacher:
https://github.com/easystats/easystats/issues/294: not sure if `.runThisTest` should be replaced by `skip_on_cran()` (what I did so far)

https://github.com/easystats/easystats/issues/363


IndrajeetPatil:
> (not sure if `.runThisTest` should be replaced by `skip_on_cran()` (what I did so far)

Yes! I think the intent was to skip the tests on CRAN and run only on devel versions of GitHub.

etiennebacher:
I really have no clue anymore why R CMD check random fails, it's all green for me.

Also, the dataset `dat.bcg` in `metafor` moved to `metadat` but it requires R >= 4.0.0 so it fails on older checks, @IndrajeetPatil where do I need to add this exception?

etiennebacher:
> I really have no clue anymore why R CMD check fails, it's all green for me.

Actually `devtools::test()` passes but tests fail in `devtools::check()`

> Fixed in https://github.com/easystats/workflows/commit/7cf7cbb1e11dc899140a8ff74d863860fef4a2bf.

Thanks!

etiennebacher:
Actually the problem is that `check_outliers(mtcars, method = c("zscore", "iqr"))` returns no outlier, so that the test is wrong but which part? The expectation or the function? @rempsyc

rempsyc:
That's my bad. I made a code change in #544. I'll fix it.

IndrajeetPatil (approved):
Thanks a ton, @etiennebacher! 💕 

That was a lot of work and we really appreciate it.

PR #574, opened 2023-04-08
Draf check_model vignette

strengejacke:
Fixes #395

PR #575, opened 2023-04-09
Vignette

strengejacke:


strengejacke:
Maybe we can integrate with #574

strengejacke:
What is not quite clear to me is the demonstration of the relationship between p and pd? Your introduction ends with

> Many of these assumptions can be checked using the `check_model()` function. We will demonstrate this function using the `mtcars` dataset.

The reader now knows why model diagnostic is important, why eye-balling is preferred over statistical tests, which kind of assumptions are important to test, and how to do this (using `check_model()`).

But then you start with simulations to show the relation between p and pd, and that interpreting p-values is not easy. Why this section? This is not directly related to model diagnostics (regarding show cases of `check_model()`).

Issue #607, opened 2023-09-10
Errors in CRAN checks


strengejacke:
@rempsyc Not sure why these tests suddenly fail:
https://cran.r-project.org/web/checks/check_results_performance.html

This commit should fix it, i.e. it should no longer error when `NULL` is returned: https://github.com/easystats/performance/commit/7b181e128215e395d09976053feb9bb8753fb137

Need to fix tests though, because these tests now would return `NULL` on CRAN:

(https://github.com/easystats/performance/blob/7b181e128215e395d09976053feb9bb8753fb137/tests/testthat/test-check_outliers.R#L100-L112)

Maybe we just skip them on CRAN, and I wonder why we need `rstan` here?

rempsyc:
Hum, strange, and they don't seem to fail on all platforms. For `rstan`, it might be because at the time, I copied that line everywhere we set a seed, but it seems it is not necessary (tried running it after uninstalling `rstan`). For the seed, it does not even seem necessary locally, but I think during CI testing not having the seed was something throwing errors, so maybe there's some randomness involved somewhere (we could try to remove it too to see if tests still pass).

For fixing the tests for CRAN, it is a bit tricky since the output will change by platform... you already skipped on Linux, which seems to be the offending OS. So I don't think we need to skip on CRAN completely, at least for now? And thanks for fixing the NULL bug.

rempsyc:
Ok, so I have downloaded the latest R development version locally and the ICS tests are successful... even the GA on R devel checks are successful, so this is hard to debug and seems like a CRAN-specific issue. So in PR #608 I skip on CRAN as you suggested.

Actually for the github action checks, we check on R devel, but only for check-strict, which does not install suggested packages, so the relevant tests are skipped anyway right? But then why don't we check for R devel for our regular R CMD check? Seems like only release or old-release are checked, so that does not allow us to compare easily with CRAN results.

strengejacke:
Thanks a lot! I think this should resolve the CRAN issues.

PR #608, opened 2023-09-10
Fix `check_outliers` ics which test on R devel

rempsyc:
Test `check_outliers` ics which on R devel re #607


strengejacke:
LGTM, thanks!

PR #616, opened 2023-09-15
try to reduce warnings in examples and use examplesIf

strengejacke:
Ref:
- https://github.com/easystats/easystats/issues/343
- https://github.com/easystats/easystats/issues/318


strengejacke:
image

Issue #632, opened 2023-10-01
double-checking `check_zeroinflation`


bbolker:
`insight::model_info` appears to have a [fairly permissive definition of is_count](https://github.com/easystats/insight/blob/443c5880bfac1170c81acd9dff94d638b67d9bcc/R/utils_model_info.R#L30-L45); in particular, it includes the generalized Poisson, `genpois` (although not `compois`); `is_negbin` also includes the generalized Poisson,

within `check_zeroinflation`, though, it seems that we assume that a model that `is_count && !is_negbin` is Poisson, and a model that is `is_negbin` is negative binomial (i.e., we compute `dpois()` or `dnbinom()`. This seems problematic:

- an `nbinom1` or `genpois` model won't have the same distribution as `dnbinom()`
- if someone tests a model that is *already zero-inflated* (this is a silly thing to do but someone could try it?) it won't give the right distribution ...

strengejacke:
Two questions:

- is `check_zeroinflation()` useful / meaningful for `nbinom1` or `genpois`? If yes, we should find a package that computes such distributions (maybe _extraDistr_?)
- should the function error for models that include a ZI component?

strengejacke:
_bump_

strengejacke:
@bwiernik any suggestions which model families to include/exclude when we speak of "count" models? When should `insight::model_info()$is_count` return `TRUE` or `FALSE`?

bwiernik:
I'm not too familiar with what check_zeroinflation() does under the hood to comment on the broader issues raised here ATM. 

For count models, Poisson, negative binomial, geometric, and binomial (not bernoulli) and their  variations would all be "count" models. Multinomial as well, though that has a somewhat different structure. Really most discrete families are count models.

strengejacke:
fixed in #643

Issue #613, opened 2023-09-14
`check_normality()` does not support models of class `glmmTMB`?


SchmidtPaul:
Is this right?

``` r
library(glmmTMB)
library(performance)

dat <- lme4::sleepstudy
m <- glmmTMB(Reaction ~ 1 + Days + (1 | Subject), data = dat)
check_normality(m)
#> `check_normality()` does not support models of class `glmmTMB`.
#> Error in attr(p.val, "data") <- x: Versuch ein Attribut von NULL zu setzen
```
<sup>Created on 2023-09-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I would have just accepted this, but

1. The Error in the second line seems somewhat unintentional
2. Other functions like `check_singularity()`, `check_overdispersion()`, `check_heteroscedasticity()` and `check_homogeneity()` *do* work with objects of class `glmmTMB` and there does not seem to be an overview of which function can deal with which classes (?).
3. I don't understand what's going on, but "glmmTMB" is mentioned within the code for `check_normality()` in lines [195](https://github.com/easystats/performance/blob/7ea53ed597b610c13bf17e2c33642271fad9ef34/R/check_normality.R#L195) and [236](https://github.com/easystats/performance/blob/7ea53ed597b610c13bf17e2c33642271fad9ef34/R/check_normality.R#L236)

<details>
<summary>Session Info</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.1 (2023-06-16 ucrt)
#>  os       Windows 10 x64 (build 19045)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  German_Germany.utf8
#>  ctype    German_Germany.utf8
#>  tz       Europe/Berlin
#>  date     2023-09-14
#>  pandoc   3.1.1 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  ! package      * version    date (UTC) lib source
#>    boot           1.3-28.1   2022-11-22 [2] CRAN (R 4.3.1)
#>    cli            3.6.1      2023-03-23 [1] CRAN (R 4.3.0)
#>    codetools      0.2-19     2023-02-01 [2] CRAN (R 4.3.1)
[...]

Issue #501, opened 2022-10-23
`check_model` check notmality in GLMs?


mattansb:
Is there a reason for this? What am I missing...

``` r
library(performance)

set.seed(3)
mu <- rpois(500, lambda = 3)
x <- rnorm(500, mu, mu*3) |> ceiling() |> pmax(0)

quine.nb1 <- MASS::glm.nb(x ~ mu)

check_normality(quine.nb1)
#> Checking normality of residuals is only useful an appropriate assumption
#>   for linear models.
#> NULL
```

But…

``` r
check_model(quine.nb1)
```

![](https://i.imgur.com/fY2qVYP.png)

<sup>Created on 2022-10-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
the current normality check isn't really appropriate generally except for linear models. something like the DhARMA QQ plots would be more appropriate

strengejacke:
_MASS_ package provides a QQ plot (with `plot()`), using std. Pearson resid. Maybe we fall back to this plot for now?

strengejacke:
Wait, we're already doing this:

``` r
library(performance)
set.seed(3)
mu <- rpois(500, lambda = 3)
x <- rnorm(500, mu, mu*3) |> ceiling() |> pmax(0)

quine.nb1 <- MASS::glm.nb(x ~ mu)
check_model(quine.nb1, check = "qq")
```

![](https://i.imgur.com/mnFNWfK.png)

``` r
plot(quine.nb1)
```

![](https://i.imgur.com/n42GAdt.png)

mattansb:
> MASS package provides a QQ plot (with plot()), using std. Pearson resid. Maybe we fall back to this plot for now?

Aren't these also irrelevant?

bbolker:
[...]

Issue #500, opened 2022-10-23
Unnecessary checks in `check_model` for NB models


mattansb:
Dispersion plot does not seem to account for the theta parameter in the model?

``` r
library(performance)

set.seed(3)
mu <- rpois(500, lambda = 3)
x <- rnorm(500, mu, mu*3) |> ceiling() |> pmax(0)

quine.nb1 <- MASS::glm.nb(x ~ mu)


check_model(quine.nb1)
```

![](https://i.imgur.com/ltKRdKo.png)

<sup>Created on 2022-10-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
@bwiernik ?

strengejacke:
Not sure, do you still think the plot is inaccurate? The results from `simulate_residuals()` (DHARMa) look in line with the plot?

``` r
library(performance)

set.seed(3)
mu <- rpois(500, lambda = 3)
x <- rnorm(500, mu, mu * 3) |>
  ceiling() |>
  pmax(0)
m <- MASS::glm.nb(x ~ mu)

check_overdispersion(simulate_residuals(m))
#> # Overdispersion test
#> 
#>  dispersion ratio =   2.442
#>           p-value = < 0.001
#> Overdispersion detected.

check_model(m)
```

![](https://i.imgur.com/Hcd15De.png)

<sup>Created on 2024-03-17 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
(edit: sorry, wrong reprex, did not reinstall the pkg)

strengejacke:
Closed in #913

PR #643, opened 2023-10-26
DHARMa implementation for new `check_residuals()` function

### LINKED ISSUE #595: DHARMa implementation for new `check_residuals()` function
Before writing any code for the new `check_residuals()` function proposed in #376, I wanted to discuss the implementation we want to go with. There are two ways to go about the implementation, depending on how deeply we want to support model checks based on DHARMa’s simulated residuals in performance.

For context, below I first cover DHARMa’s basic workflow, then discuss two possible options for supporting DHARMa in performance.

The big picture questions are:

-   Which model classes do we want to support?
-   Do we want the simulated residuals to follow a uniform or normal distribution?
-   Do we want DHARMa’s simulated residuals to be supported throughout performance, or just in the new `check_residuals()` function?
    -   If we only support DHARMa’s simulated residuals in the new `check_residuals()` function then the implementation is pretty straightforward
    -   If we support DHARMa’s simulated residuals throughout performance then we’ll need to add S3 methods for objects of class `DHARMa` to various `check_()` functions (e.g., autocorrelation, zero-inflation, etc.), so the implementation will be a little more involved

### Simulated residuals

``` r
library(DHARMa)
#> This is DHARMa 0.4.6. For overview type '?DHARMa'. For recent changes, type news(package = 'DHARMa')
library(glmmTMB)
```

The basic workflow in DHARMa involves fitting a model, then simulating residuals from the fitted model.

``` r
model <- glmmTMB(
  count ~ mined + spp + (1 | site),
  family = poisson,
  data = Salamanders
)

# Note that this function sets a seed by default, so we don't need to set one
# ourselves.
simulated_residuals <- simulateResiduals(model)

simulated_residuals
#> Object of Class DHARMa with simulated residuals based on 250 simulations with refit = FALSE . See ?DHARMa::simulateResiduals for help. 
#>  
#> Scaled residual values: 0.1555024 0.734551 0.4477616 0.5050258 0.4289974 0.2123359 0.2858764 0.5904198 0.8035344 0.3373841 0.5761363 0.1596766 0.2655446 0.04189079 0.147868 0.2487158 0.5633855 0.364279 0.4414209 0.4806934 ...
```

Regarding the fitted model, DHARMa currently supports the following classes:

> Currently supported are linear and generalized linear (mixed) models from ‘lme4’ (classes ‘lmerMod’, ‘glmerMod’), ‘glmmTMB’, ‘GLMMadaptive’ and ‘spaMM’, generalized additive models (‘gam’ from ‘mgcv’), ‘glm’ (including ‘negbin’ from ‘MASS’, but excluding quasi-distributions) and ‘lm’ model classes

The main interest in the `check_residuals()` function is to add better checks for GL(M)Ms, so do we want to add methods for all these classes, or exclude some (e.g., `lm`)?

The `simulated_residuals` object is of class `DHARMa` and is structured like so. Note that the fitted model is stored as the first item in the list here, so the `simulated_residuals` object could be passed to other methods in performance that are based on a model object as input if methods are added to support the `DHARMa` class.

``` r
str(simulated_residuals, max.level = 1)
#> List of 17
#>  $ fittedModel            :List of 7
#>   ..- attr(*, "class")= chr "glmmTMB"
#>  $ modelClass             : chr "glmmTMB"
#>  $ additionalParameters   : list()
#>  $ nObs                   : int 644
#>  $ nSim                   : num 250
#>  $ refit                  : logi FALSE
#>  $ observedResponse       : int [1:644] 0 0 0 2 2 1 1 2 4 1 ...
#>  $ integerResponse        : logi TRUE
#>  $ problems               : list()
#>  $ fittedPredictedResponse: num [1:644] 0.197 0.197 0.197 1.896 1.896 ...
#>  $ fittedFixedEffects     : Named num [1:8] -1.625 2.264 -1.386 0.231 -0.77 ...
#>   ..- attr(*, "names")= chr [1:8] "(Intercept)" "minedno" "sppPR" "sppDM" ...
#>  $ fittedResiduals        : Named num [1:644] -0.2139 -0.7867 -0.0828 -0.2951 -0.2951 ...
#>   ..- attr(*, "names")= chr [1:644] "1" "2" "3" "4" ...
#>  $ method                 : chr [1:2] "PIT" "traditional"
#>  $ simulatedResponse      : num [1:644, 1:250] 0 0 0 0 3 1 2 2 1 1 ...
#>   ..- attr(*, "dimnames")=List of 2
#>  $ scaledResiduals        : num [1:644] 0.156 0.735 0.448 0.505 0.429 ...
#>  $ time                   : 'proc_time' Named num [1:5] 0.149 0.009 0.159 0 0
#>   ..- attr(*, "names")= chr [1:5] "user.self" "sys.self" "elapsed" "user.child" ...
#>  $ randomState            :List of 4
#>  - attr(*, "class")= chr "DHARMa"
```

To get the residuals stored in `simulated_residuals` use `residuals()`.

``` r
simulated_residuals |>
  residuals() |>
  head(20)
#>  [1] 0.15550238 0.73455098 0.44776164 0.50502581 0.42899743 0.21233592
#>  [7] 0.28587644 0.59041980 0.80353437 0.33738415 0.57613629 0.15967657
#> [13] 0.26554460 0.04189079 0.14786802 0.24871584 0.56338550 0.36427899
#> [19] 0.44142089 0.48069339
```

By default the simulated residuals have a uniform distribution, but the scaling can be transformed to any other distribution, such as the normal distribution. However, in the case of the normal distribution, outliers will be mapped to -Inf / Inf by default (and if you want to get around this you need to decide which lower and upper values to convert -Inf / Inf to). From the [DHARMa vignette](https://cran.r-project.org/web/packages/DHARMa/vignettes/DHARMa.html#workflow-in-dharma)

> These normal residuals will behave exactly like the residuals of a linear regression. However, for reasons of a) numeric stability with low number of simulations, which makes it neccessary to descide on which value outliers are to be transformed and b) my conviction that it is much easier to visually detect deviations from uniformity than normality, DHARMa checks all residuals in the uniform space, and I would personally advice against using the transformation.

``` r
# Note the Inf values here:
simulated_residuals |>
  residuals(quantileFunction = qnorm) |>
  tail(30)
#>  [1]  1.08007805  0.10778534  0.09064290 -0.25809108         Inf -2.00160575
#>  [7]         Inf -1.09151823  0.79929750  0.57712362 -0.97183120 -0.22312363
#> [13] -0.25864418  1.31949100 -1.43237801  1.07378421 -0.29403004 -0.20597054
#> [19]  1.15364152  2.09225482 -0.02295151 -0.80227605 -0.31481034  0.20750053
#> [25] -1.12027563 -0.97722105  1.01980584  1.85693880 -1.51072945         Inf

# Convert -Inf / Inf to -7 / 7:
simulated_residuals |>
  residuals(quantileFunction = qnorm, outlierValues = c(-7, 7)) |>
  tail(30)
#>  [1]  1.08007805  0.10778534  0.09064290 -0.25809108  7.00000000 -2.00160575
#>  [7]  7.00000000 -1.09151823  0.79929750  0.57712362 -0.97183120 -0.22312363
#> [13] -0.25864418  1.31949100 -1.43237801  1.07378421 -0.29403004 -0.20597054
#> [19]  1.15364152  2.09225482 -0.02295151 -0.80227605 -0.31481034  0.20750053
#> [25] -1.12027563 -0.97722105  1.01980584  1.85693880 -1.51072945  7.00000000
```

Given this outlier mapping issue, I think it’s probably best to stick with the uniform distribution unless there’s an opinionated way to automatically decide the range.

The main reason I brought this up though was that if transformed to normal, then we can use standard residual checks that assume normal residuals instead of the tests shipped with DHARMa; so some of the existing methods in performance could be reused here (but the sticky issue of defining the range for outliers would have to be dealt with).

### Residual tests

DHARMa provides a number of specialized goodness-of-fit tests on the simulated residuals:

-   `testUniformity()`: tests if the overall distribution conforms to expectations
-   `testOutliers()`: tests if there are more simulation outliers than expected
-   `testDispersion()`: tests if the simulated dispersion is equal to the observed dispersion
-   `testQuantiles()`: fits a quantile regression or residuals against a predictor (default predicted value), and tests of this conforms to the expected quantile
-   `testCategorical(simulationOutput, catPred = testData$group)` tests residuals against a categorical predictor
-   `testZeroinflation`(): tests if there are more zeros in the data than expected from the simulations
-   `testGeneric`(): test if a generic summary statistics (user-defined) deviates from model expectations
-   `testTemporalAutocorrelation`(): tests for temporal autocorrelation in the residuals
-   `testSpatialAutocorrelation`(): tests for spatial autocorrelation in the residuals. Can also be used with a generic distance function, for example to test for phylogenetic signal in the residuals

The proposed `performance::check_residuals()` function could either use `DHARMa::testUniformity()` under the hood:

``` r
uniformity <- testUniformity(simulated_residuals, plot = FALSE)

class(uniformity)
#> [1] "ks.test" "htest"

uniformity
#> 
#>  Asymptotic one-sample Kolmogorov-Smirnov test
#> 
#> data:  simulationOutput$scaledResiduals
#> D = 0.086753, p-value = 0.0001234
#> alternative hypothesis: two-sided
```

Or just use the Kolmogorov-Smirnov test directly, since `DHARMa::testUniformity()` is basically just a wrapper function with an additional argument for calling a plot as well.

``` r
ks.test(residuals(simulated_residuals), "punif")
#> Warning in ks.test.default(residuals(simulated_residuals), "punif"): ties should
#> not be present for the Kolmogorov-Smirnov test
#> 
#>  Asymptotic one-sample Kolmogorov-Smirnov test
#> 
#> data:  residuals(simulated_residuals)
#> D = 0.086753, p-value = 0.0001234
#> alternative hypothesis: two-sided
```

## Basic DHARMa implementation

A basic implementation of `check_residuals()` (ignoring the S3 methods for now) could do something like this, where the residuals are simulated, the test is run and printed, then a list containing the test results and residuals is returned.

``` r
check_residuals <- function(x) {
  simulated_residuals <- residuals(simulateResiduals(model))
  test_results <- suppressWarnings(ks.test(simulated_residuals, "punif"))
  # This would be printed nicely in the actual implementation, but you get the
  # idea.
  print(test_results)
  # We need to store the residuals here so they can be passed to the plot()
  # method that will be added to `see` later.
  invisible(list(test = test_results, residuals = simulated_residuals))
}

check_residuals(model)
#> 
#>  Asymptotic one-sample Kolmogorov-Smirnov test
#> 
#> data:  simulated_residuals
#> D = 0.086753, p-value = 0.0001234
#> alternative hypothesis: two-sided
```

## Involved DHARMa implementation

A more involved implementation would involve us adding a `simulate_residuals()` function that’s a wrapper of `DHARMa::simulateResiduals()`. We would give this function S3 methods for the different model classes we want to support.

This object could then be passed to the various `check_()` functions we want to support. For example:

-   `check_residuals.DHARMa()`
-   `check_autocorrelation.DHARMa()`
-   `check_overdispersion.DHARMa()`
-   etc.

If we went with this approach, we’d advocate for a workflow like the basic DHARMa workflow demonstrated above. Since the `simulated_residuals()` object also contains the fitted model, this could also be easily passed onto `check_model()`.

<sup>Created on 2023-06-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #595

I'm not sure, but I think this PR _could_ also be related to following issues:
- [x] #688 
- [x] #654
- [x] #501
- [x] #500
- [ ] #464
- [x] #632
- [x] #613
- [x] #408
- [x] #405
- [x] #376
- [x] #367
- [ ] #274
- [x] #263

**To do**
- [x] `check_overdispersion()` based on `simulated_residuals()`
- [ ] `plot()` for `check_overdispersion()`


strengejacke:
@mccarthy-m-g I invited you to get write access, which should allow you to push to this PR, so we can collaboratively work on this.

mccarthy-m-g:
Thanks! Accepted.

It will probably be a month or two before I can dedicate time to this, but I'll look at the issues you linked above today.

strengejacke:
I don't think you need to look into all those issues I linked to, that was more a reminder for myself ;-) We should focus on #595, and then I'll check to which extent other issues might be resolved.

mccarthy-m-g:
Sounds good, although I already started looking... 😅 (DHARMa doesn't support objects of class "lrm", so you can drop #471 from consideration)

I think the `performance::check_model.performance_simres` method that will be added could probably address some of the issues indirectly (e.g., #501), but that could be dealt with after this PR is merged.

strengejacke:
> but that could be dealt with after this PR is merged

Agreed.

strengejacke:
So the basic workflow would be:

1. `out <- simulate_residuals(model)`
2. `check_model(out)` or `check_<whatever>(out)`?

Then we wouldn't need particular `plot()` or `print()` methods, except for some additional functions (like the suggested `check_residuals()`).

mccarthy-m-g:
> So the basic workflow would be:
> 
> 1. `out <- simulate_residuals(model)`
> 2. `check_model(out)` or `check_<whatever>(out)`?
> 
> Then we wouldn't need particular `plot()` or `print()` methods, except for some additional functions (like the suggested `check_residuals()`).

Yes and no. That's the basic workflow, but we'll want a print method for `performance_simres` and the `check_()` functions (at least under my vision for how this should get implemented), and we don't want to return the residuals directly in `simulate_residuals()`

I'll add some rough commits with comments for context.

mccarthy-m-g:
If you look at the changes from those commits, particularly the vignette I added, does my rationale for why we don't want to prematurely return residuals in `simulate_residuals()` make sense?

strengejacke:
> If you look at the changes from those commits, particularly the vignette I added, does my rationale for why we don't want to prematurely return residuals in `simulate_residuals()` make sense?

I see, yes, makes sense!

I'm note sure how to align the `plot()` method. For testing purposes, I opened a PR (https://github.com/easystats/see/pull/312), but the plots look rather different.

``` r
library(performance)
library(glmmTMB)

model <- glmmTMB(
  count ~ mined + spp + (1 | site),
  family = poisson,
  data = Salamanders
)

res <- simulate_residuals(model)
x <- check_normality(res)
plot(x)
```



``` r
plot(DHARMa::simulateResiduals(model))
#> DHARMa:testOutliers with type = binomial may have inflated Type I error rates for integer-valued distributions. To get a more exact result, it is recommended to re-run testOutliers with type = 'bootstrap'. See ?testOutliers for details
```



<sup>Created on 2023-10-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mccarthy-m-g:
> I'm note sure how to align the plot() method. For testing purposes, I opened a PR, but the plots look rather different.

DHARMa uses a uniform distribution, *not* a normal distribution, for its left plot. CRAN is down right now and I don't have qqplotr installed so I can't share a reprex, but this code should be what we want as the plot method for `check_residuals()`:

```r
library(glmmTMB)

model <- glmmTMB(
  count ~ mined + spp + (1 | site),
  family = poisson,
  data = Salamanders
)

simulated_residuals <- simulate_residuals(model)

plot_simulated_residuals <- function(x) {

  dp <- list(min = 0, max = 1, lower.tail = TRUE, log.p = FALSE)
  ggplot2::ggplot(
    tibble::tibble(scaled_residuals = residuals(x)),
    ggplot2::aes(sample = scaled_residuals)
  ) +
    qqplotr::stat_qq_band(distribution = "unif", dparams = list(min = 0, max = 1), alpha = .2) +
    qqplotr::stat_qq_line(distribution = "unif", dparams = dp, size = .8, colour = "#3aaf85") +
    qqplotr::stat_qq_point(distribution = "unif", dparams = dp, size = .5, alpha = .8, colour = "#1b6ca8") +
    ggplot2::labs(
      title = "Uniformity of Residuals",
      subtitle = "Dots should fall along the line",
      x = "Standard Uniform Distribution Quantiles",
      y = "Sample Quantiles"
    ) +
    see::theme_lucid()
}

plot_simulated_residuals(simulated_residuals)
```

I'm a bit skeptical about having a `check_normality()` method here too, unless we restrict it to the models DHARMa supports where that would make sense to do (e.g., `lm`). Otherwise it might be better to just return a message and `NULL` value pointing the user to the `check_residuals()` function.

bwiernik:
I would suggest we do `check_residuals(model)` as the workflow so that we can have correct titles for the plot instead of "Normality of residuals"

strengejacke:
ok, so `check_residuals()` would be an additional check we add to the existing functions, right? What about `check_model()`? Which residual plots do we want there? Or should we add the `check_residuals()` plot as additional panel? But then it might be confusing when we have different plots checking for "normality" of residuals (i.e. we have two different qq-plots).

mccarthy-m-g:
> ok, so check_residuals() would be an additional check we add to the existing functions, right?

Yeah. We could discuss using `check_normality()` on simulated residuals later (since I think these residuals can be used for that for models where normality checks are a thing), but for now the main focus should be the `check_residuals()` implementation since that’s the main feature add for this PR.

> What about check_model()? Which residual plots do we want there?

My idea for this was that you would run `check_model()` on the object returned by `simulate_residuals()` rather than the model object itself, and the residuals plot would be the uniform distribution one DHARMa uses, with an appropriate title that doesn’t mention “normality” anywhere. So using simulated residuals would be more of an opt-in workflow than an opinionated default. I think this is a good starting point for the PR.

After that basic workflow is implemented I think it would be easier to figure out whether or not this should be the only way to get simulated residual plots, or if they should be also be available through something like `check_model(model, residual_type = “simulated”)`.

bwiernik:
> ok, so `check_residuals()` would be an additional check we add to the existing functions, right? What about `check_model()`? Which residual plots do we want there? Or should we add the `check_residuals()` plot as additional panel? But then it might be confusing when we have different plots checking for "normality" of residuals (i.e. we have two different qq-plots).

For non-normal Gaussian models, we would replace the normality and plot with the residuals plot

strengejacke:
``` r
library(performance)
library(glmmTMB)
#> Warning in checkDepPackageVersion(dep_pkg = "TMB"): Package version inconsistency detected.
#> glmmTMB was built with TMB version 1.9.9
#> Current TMB version is 1.9.10
#> Please re-install glmmTMB from source or restore original 'TMB' package (see '?reinstalling' for more information)

model <- glmmTMB(
  count ~ mined + spp + (1 | site),
  family = poisson,
  data = Salamanders
)

check_model(model)
#> `check_outliers()` does not yet support models of class `glmmTMB`.
```



``` r
check_model(model, detrend = FALSE)
#> `check_outliers()` does not yet support models of class `glmmTMB`.
```



<sup>Created on 2024-03-15 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
``` r
library(performance)
library(glmmTMB)
#> Warning in checkDepPackageVersion(dep_pkg = "TMB"): Package version inconsistency detected.
#> glmmTMB was built with TMB version 1.9.9
#> Current TMB version is 1.9.10
#> Please re-install glmmTMB from source or restore original 'TMB' package (see '?reinstalling' for more information)

model <- glmmTMB(
  count ~ mined + spp + (1 | site),
  family = poisson,
  data = Salamanders
)

plot(simulate_residuals(model))
```



``` r
plot(simulate_residuals(model), detrend = TRUE)
```



<sup>Created on 2024-03-15 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
@bwiernik and @mccarthy-m-g - I think the implementation works quite well now for the first methods. The `plot()` method works fine for the Q-Q plots (https://github.com/easystats/see/pull/329). Things we should consider:

- We would have to update the plot for overdispersion/zero-inflation checks (see https://github.com/easystats/performance/pull/643#issuecomment-1999544659). This is still based on the classical residuals, not the simulated ones. I have played around with revising the current code, copying the function `.diag_overdispersion` into a new `.new_diag_overdispersion` (see https://github.com/easystats/performance/blob/ad60db15550c406e533ad037107c81d81a66590d/R/check_model_diagnostics.R#L296). This did not really work - do you have any ideas how we can have new plots for overdispersion/zero-inflation? I found it quite informative.

- `check_zeroinflation()` and `check_overdispersion()` now rely on `simulate_residuals()` for zero-inflated or negative binomial models etc., so only the really "simple" models that returned identical results to DHARMa-tests still use the old code. Only Poisson mixed models also use the "old" code, which still could be inaccurate (see #464) - do we want to use `simulate_residuals()` in general for mixed models when calling `check_zeroinflation()` and `check_overdispersion()`?

strengejacke:
Here's the implementation for `check_outliers()`:

``` r
library(performance)

# For statistical models ---------------------------------------------
# select only mpg and disp (continuous)
mt1 <- mtcars[, c(1, 3, 4)]
# create some fake outliers and attach outliers to main df
mt2 <- rbind(mt1, data.frame(
  mpg = c(37, 40), disp = c(300, 400),
  hp = c(110, 120)
))
# fit model with outliers
model <- lm(disp ~ mpg + hp, data = mt2)

res <- simulate_residuals(model)
check_outliers(res)
#> # Outliers detection
#> 
#>   Proportion of observed outliers: 2.94%
#>   Proportion of expected outliers: 0.80%, 95% CI [0.07, 15.33]
#> No outliers were detected (p = 0.238).
DHARMa::testOutliers(res, plot = FALSE)
#> 
#>  DHARMa outlier test based on exact binomial test with approximate
#>  expectations
#> 
#> data:  res
#> outliers at both margin(s) = 1, observations = 34, p-value = 0.2381
#> alternative hypothesis: true probability of success is not equal to 0.007968127
#> 95 percent confidence interval:
#>  0.0007443642 0.1532676696
#> sample estimates:
#> frequency of outliers (expected: 0.00796812749003984 ) 
#>                                             0.02941176

DHARMa::testOutliers(res, type = "bootstrap", plot = FALSE)
#> 
#>  DHARMa bootstrapped outlier test
#> 
#> data:  res
#> outliers at both margin(s) = 1, observations = 34, p-value = 0.78
#> alternative hypothesis: two.sided
#>  percent confidence interval:
#>  0.00000000 0.05882353
#> sample estimates:
#> outlier frequency (expected: 0.015 ) 
#>                           0.02941176
check_outliers(res, type = "bootstrap", iterations = 200)
#> # Outliers detection
#> 
#>   Proportion of observed outliers: 2.94%
#>   Proportion of expected outliers: 1.32%, 95% CI [0.00, 5.88]
#> No outliers were detected (p = 0.690).
```

<sup>Created on 2024-03-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
The more I dive into the DHARMa topic, the more I realize that this approach is quite similar to what we aimed at with `check_predictions()`.

Issue #202, opened 2021-02-22
Reproduce manually rstan's pp_check


DominiqueMakowski:
Opening this for myself, will close soon

``` r
m <- rstanarm::stan_glm(mpg ~ am, data=mtcars, refresh=0)
rstanarm::pp_check(m)
```

![](https://i.imgur.com/uCRRqYN.png)

<sup>Created on 2021-02-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

- **get predicted**: `pred <- as.data.frame(insight::get_predicted(m))`

strengejacke:
> Wait no but actually I want to replicate the rstanarm thing for Bayesian models:

Yeah, but I did this for freq. models, because for Bayesian models it's already available? 

> For Bayesian models, pp_check is from stan, but the other one would return our version (rather than the one using the default aesthetic of rstan)

What would be the difference?

DominiqueMakowski:
our version is more beautiful? 😁 + we can add more info like a title? + it gives a unified output for freq and bayesian (same colors, etc)

strengejacke:
ok, got it.

DominiqueMakowski:
Also, our pp_check is currently not defined for brmsfit objects :(, and for complex models the brms/rstantools version fails:

<details>
  <summary>reprex</summary>
  
``` r
library(brms)
#> Loading required package: Rcpp
#> Loading 'brms' package (version 2.15.0). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:stats':
#> 
#>     ar



# Stan family -------------------------------------------------------------

rtmix <- brms::custom_family(
  "rtmix",
  dpars = c("mu", "sigma", "mix", "shiftprop"), # Those will be estimated
  links = c("identity", "log", "logit", "logit"),
  type = "real",
  lb = c(NA, 0, 0, 0), # bounds for the parameters
  ub = c(NA, NA, 1, 1),
  vars = c("vreal1[n]", "vreal2[n]") # Data for max_shift and upper (known)
)

# Stan model -------------------------------------------------------------------

stan_functions <- brms::stanvar(block = "functions", scode = "
  real rtmix_lpdf(real y,
                      real mu,
                      real sigma,
                      real mix,
                      real shiftprop,
                      real max_shift,
                      real upper) {
    real shift = shiftprop * max_shift;
[...]

PR #706, opened 2024-03-29
Remove `pp_check()` alias

### LINKED ISSUE #477: Remove `pp_check()` alias
We had kept this for backward compatibility, but given its collision with `{bayesplot}`, I don't think it's a good idea to continue to retain it.

`posterior_predictive_check()` is good enough.

strengejacke:
Fixes #477


R/check_predictions.R
```diff
@@ -161,10 +161,37 @@ check_predictions.stanreg <- function(object,
   )
 
   if (inherits(object, "brmsfit")) {
-    out <- bayesplot::pp_check(object, type = type, ndraws = iterations, ...)
+    out <- as.data.frame(bayesplot::pp_check(object, type = type, ndraws = iterations, ...))
```
DominiqueMakowski: do we even need to call `bayesplot`? Couldn't we directly compute predictions like in https://github.com/easystats/performance/issues/202#issuecomment-783842658

though it might be less efficient

strengejacke: Not sure how we can get dot-plots for discrete/integer outcomes?

bwiernik:
Why do we want to remove the aliases? I thought we had resolved the conflicts with pp_check(), and think there is value in exporting methods for that generic?

strengejacke:
We don't want to remove `pp_check()`, only `posterior_predictive_check()` and `check_posterior_predictions()`.

PR #764, opened 2024-08-19
Adding `r2_mlm()` and possible update to `r2()` to work with it

### LINKED ISSUE #758: Proposed Multivariate $R^2$: `r2_mlm` for "mlm" Classed Model Objects
At current, the `r2()` method for model objects classed `c("mlm","lm")` returns $R^2$ values for both outcomes separately and describes the proportion of variance in each of the separate outcomes.

There are some applications, in particular for relative importance analysis/dominance analysis, in which it is valuable to obtain one $R^2$ value that integrates across outcomes and provides a single value that describes the proportion of variance explained in the system of outcomes by the predictors.

Two recommended $R^2$ metrics are discussed by Azen and Budescu (2006) for use in dominance analysis that are derived from Van den Berg and Lewis (1988).

Would propose to add a `r2_mlm()` function that calls both of the recommended $R^2$ metrics for "mlm" classed model objects. I would also be happy to implement this function and submit a pull request if the maintainers are open to this additional set of metrics.

### References

Azen, R., & Budescu, D. V. (2006). Comparing predictors in multivariate regression models: An extension of dominance analysis. _Journal of Educational and Behavioral Statistics, 31(2),_ 157-180.

Van den Burg, W., & Lewis, C. (1988). Some properties of two measures of multivariate association. _Psychometrika, 53,_ 109-122.

jluchman:
Fixes #758

This pull request includes a new function `r2_mlm()` that generates two $R^2$ values for multivariate linear regression objects/objects of class `c("mlm", "lm")` and is associated with issue #758.

This pull request also adds an `multivariate` logical argument to the `r2()` function for its "mlm" S3 method that can invoke the computation of the `r2_mlm` but is by default 'off'/`FALSE` and returns the current $R^2$ metric separated by response. I see value in this addition to the `r2()` function but do understand if it is not in the spirit of the function and it is not deemed a value add.


DominiqueMakowski:
what I consider is the opposite: running `r2(mlm)` would return the multivariate R2 (i.e., would run `r2_mlm()`), but there would be an argument (documented in `r2_mlm()`) to set `multivariate=FALSE` (true by default) that would return the r2s per response.

jluchman:
Update just pushed that incorporates the `multivariate=TRUE` default for `r2(mlm)` along with a few edits made by the linter and some minor documentation updates.

strengejacke (approved):

strengejacke:
Thanks a lot, looks good to me!
Will wait for checks and then merge.

PR #777, opened 2024-11-14
[docs] Details to performance_roc()

DominiqueMakowski:
There is probably a better way to phrase it though (@mattansb?)


mattansb:
The docs look great (:

DominiqueMakowski:
why does the `as.numeric()` method fail?? It fails locally too, did  I forgot how to R?

strengejacke:
For me, it also fails locally and I couldn't find any mistake? Strange, what have we missed?

strengejacke:
`performance:::as.numeric.performance_roc(x)` works, `as.numeric(x)` not. @etiennebacher, any ideas, why the `as.numeric.performance_roc()` method in this PR does not work, resp. is not called?

``` r
library(performance)
library(bayestestR)
data(iris)

set.seed(123)
iris$y <- rbinom(nrow(iris), size = 1, .3)
folds <- sample(nrow(iris), size = nrow(iris) / 8, replace = FALSE)
test_data <- iris[folds, ]
train_data <- iris[-folds, ]

model <- glm(y ~ Sepal.Length + Sepal.Width, data = train_data, family = "binomial")
x <- performance_roc(model, new_data = test_data)

class(x)
#> [1] "performance_roc"     "see_performance_roc" "data.frame"

as.numeric(x)
#> Error: 'list' object cannot be coerced to type 'double'

performance:::as.numeric.performance_roc(x)
#> [1] 0.3766234
```

<sup>Created on 2024-12-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher:
You need to add the S3 method for `as.double.performance_roc` instead of  `as.numeric.performance_roc`. From `?as.numeric`:

> as.numeric is a generic function, but S3 methods must be written for as.double. It is identical to as.double.

R/performance_roc.R
```diff
@@ -122,6 +123,21 @@ print.performance_roc <- function(x, ...) {
...
+  } else {
+    dat <- split(x, f = x$Model)
+
+    auc <- c()
+    for (i in seq_along(dat)) {
+      auc <- c(auc, bayestestR::area_under_curve(dat[[i]]$Specificity, dat[[i]]$Sensitivity))
+    }
+  }
+  auc
+}
```
etiennebacher: I don't know if you expect many models to be passed in `x` but if so, it will be more performant to preallocate the vector instead of extending it in the loop:

```r
auc <- numeric(length(dat))
for (i ...) {
  auc[i] <- bayestestR::...
}
```

strengejacke:
Coo, thanks!

PR #785, opened 2024-12-02
[Fix] check_outliers() expose 'xi' parameter for OPTICS

DominiqueMakowski:
I have a nice new dataset of a questionnaire validation that included attention checks. I have been playing with check_outliers(), and the OPTICS method. Interestingly, the current rule for its *threshold* that we have didn't work at all, + I had to change `xi`  to actually make it work (hence this PR that allows to do that).

Interestingly though, we typically consider outliers observations with a **large** distance, but my data shows that people that have failed  attention checks (i.e., "objective" outliers) tend  to have a low distance:

image

(the logistic model failed ~ distance_optics makes it a significant predictor of failure)

I wouldn't recommend any action for now, maybe we can simply mention this observation (that it is possible that low distances are also problematic) once my paper is published and link to it

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.12.4.12
+Version: 0.12.4.13
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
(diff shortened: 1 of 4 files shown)


strengejacke:
Looks good, can be merged, or not?

DominiqueMakowski:
Yup, it should work! Thanks

PR #798, opened 2025-03-12
[Feature] check_reliability() fox mixed models (random effects variability-over-uncertainty)

DominiqueMakowski:
Now that modelbased' estimate_grouplevel() is robust, we can leverage that to "automatically" compute reliability based on the approach outlined in a couple of recent papers. 

@anafneves @emmalbenn it's data analysis software dev time! I would like you to:
1. Read carefully Rouder's papers ([this](https://osf.io/preprints/psyarxiv/8ktn6_v1), [this](https://journals.sagepub.com/doi/10.1177/09637214231220923)) and potentially others
- see also:
  - https://easystats.github.io/modelbased/articles/estimate_grouplevel.html
  - https://realitybending.github.io/post/2024-03-18-signaltonoisemixed/
  - https://osf.io/preprints/psyarxiv/2ux7t_v1
  - https://www.researchgate.net/publication/341560356_Beneath_the_surface_Unearthing_within-person_variability_and_mean_relations_with_Bayesian_mixed_models
2. Summarize and explain to me all of it
3. Check whether the computation of this reliability that he introduces matches my implementation in this function
4. Write the missing docstrings sections (currently marked by "TODO"). You can either make a PR to this PR with the changes, or the easier version is to comment here the text and I can copy and commit

You have 24h. Your time starts now 🕙


strengejacke:
> You have 24h. Your time starts now 🕙

😆 

<sub>(no pressure...)</sub>

strengejacke:
space_around

DominiqueMakowski:
look daniel life's too short to put spaces before left parentheses 😅 yolo

strengejacke:
ctrl+shift+a (after installing air extension). 😎

R/check_reliability.R
```diff
@@ -45,24 +43,24 @@ check_reliability.estimate_grouplevel <- function(x, ...) {
   if (length(unique(x$Level)) <= 3) {
     insight::format_alert(paste0(
       "The number of random levels (N = ",
-      length(unique(x$Level)),
+      insight::n_unique(x$Level),
```
DominiqueMakowski: given that it makes the code longer so I'm not sure it's necessary here 😁

strengejacke: it's more robust, as it deals with `NA` values :-)
and of course, our functions are always preferable.

DominiqueMakowski: > and of course, our functions are always preferable.

Can't argue with that

image

bwiernik:
Also check out this package that implemented formulas from the HLM book for reliability https://github.com/jadahlke/hlmer

Would this conceptually fit better as a `performance_*` function versus `check_*`?

R/check_reliability.R
strengejacke: (and it has methods for lists and data frames, too)

DominiqueMakowski:
> Would this conceptually fit better as a performance_* function versus check_*?

That's a good question I wondered  whether to add that as a check_ or performance_ so I think there might some conceptual overlap but what I thought:

- performance_* functions assess whether a model is *good* (good fit, good predictive value, etc.)
- check_* assess whether a model is viable and well-parameterized (i.e., it's about the inside of a model)

In this case, we could imagine the case that adding non-reliable random effects does improve performance. I.e., the reliability is not necessarily a function of performance (?), hence to add it as a check because it looks at something specific within the model

DominiqueMakowski:
### About hlmer

#### How Reliability is Computed in hlmer

The hlmer package calculates reliability estimates for random effects as part of its augmentation of the lmer() output from the lme4 package. These estimates are inspired by the approach used in the HLM7 software, which is designed for hierarchical linear modeling. The reliability of random effects reflects how precisely the random effects (e.g., random intercepts or slopes) are estimated for each cluster (e.g., schools, groups), given the data and model structure.

#### Explanation

Reliability in this context measures how trustworthy the estimated random effects are for each cluster. It’s akin to asking: “How much of the variability in these random effects (e.g., school-specific intercepts) is due to true differences between clusters versus noise from limited data?”
Formula Intuition: The reliability formula tau / (tau + SE^2) is a signal-to-noise ratio:
  - tau (signal): The true variance of the random effects across clusters.
  - SE^2 (noise): The sampling variance, which depends on cluster size and residual variance.
  - High reliability (close to 1): The true variance dominates, meaning the random effects are precisely estimated (large clusters, low noise).
  - Low reliability (close to 0): Noise dominates, meaning the estimates are unreliable (small clusters, high residual variance).


#### What It Does

For researchers, this statistic flags whether the random effects are meaningful. Low reliability suggests that differences between clusters might be artifacts of small sample sizes rather than real effects.

It’s especially useful in hierarchical models where clusters (e.g., schools, patients) vary in size. A school with 5 students might have a noisy intercept estimate, while one with 50 students is more stable.

### Difference from my proposal

Now, let’s compare hlmer’s reliability to your proposed alternative: extracting random effect coefficients and their standard errors, then computing a ratio of between-group variability (sd(coefficients)) to mean within-group variability (mean(SEs)).

#### Alternative Approach

Steps:
1. Extract the random effect coefficients (e.g., cluster-specific intercepts) from the model (e.g., using ranef(mod) in lme4).
2. Extract the standard errors of these random effects (possibly from conditional variance estimates or cluster-specific calculations).
3. Compute:
4. Between-group variability = sd(coefficients) (standard deviation of the random effects across clusters).
5. Mean within-group variability = mean(SEs) (average standard error of the random effects).
6. Ratio = sd(coefficients) / mean(SEs).

Intuition: This ratio compares the observed spread of random effects to their average uncertainty, suggesting how “real” the differences are relative to estimation noise.

#### Comparison to hlmer Reliability

Let’s break this down:

##### Conceptual Similarity:
Both approaches aim to quantify the signal (true variability) versus noise (estimation error) in random effects.

hlmer’s tau / (tau + SE^2) and the alternative’s sd(coefficients) / mean(SEs) are both signal-to-noise ratios in spirit.

##### Computation Differences:

- Numerator:
  - hlmer: Uses tau, the model-estimated variance of the random effects (from varcor), which is a population-level estimate.
  - Alternative: Uses sd(coefficients), the sample standard deviation of the extracted random effects, which is an empirical estimate from the fitted values.
  - Difference: tau is a model-based parameter, while sd(coefficients) reflects the realized variability in this sample, potentially shrunk toward zero by the model’s empirical Bayes estimation.
- Denominator:
  - hlmer: Uses tau + SE^2, where SE^2 is the cluster-specific sampling variance, averaged across clusters.
  - Alternative: Uses mean(SEs), the average standard error of the random effects.
  - Difference: hlmer incorporates both true variance and error variance in the denominator, while the alternative isolates the error term. Also, SE^2 in hlmer is computed analytically per cluster (via se_lvl1()), whereas SEs in the alternative might come from conditional variances or approximations.
- Form:
  - hlmer: A reliability metric (0 to 1), directly interpretable as precision.
  - Alternative: A ratio that isn’t bounded (could exceed 1), more akin to a standardized effect size.

##### Practical Differences:

- Data Dependency:
  - hlmer: Relies on model variance components and cluster sizes, computed before averaging across clusters.
  - Alternative: Depends on the extracted random effects, which are post hoc and influenced by shrinkage (random effects are conditional means, adjusted toward the grand mean when data is sparse).

- Scope:
  - hlmer: Computes reliability for each random effect (intercept, slopes) separately.
  - Alternative: Typically applied to one random effect at a time (e.g., intercepts), unless extended to slopes explicitly.

- Robustness:
  - hlmer: Uses theoretical variances, making it consistent with the model’s assumptions.
  - Alternative: Sensitive to the sample of clusters and shrinkage effects, potentially underestimating between-group variability if random effects are heavily shrunk.


##### Pros and Cons

- hlmer Reliability:
  - Pro: Theoretically grounded, aligns with HLM7, bounded (0–1), handles multiple random effects.
  - Con: Assumes model variances are accurate, less intuitive as a standalone ratio.
- Alternative Ratio:
  - Pro: Simple, uses observed estimates, flexible for post hoc analysis.
  - Con: Unbounded, affected by shrinkage, requires careful SE extraction, less tied to model assumptions.

Which is Better?
Depends on your goal:
Use hlmer reliability if you want a standardized, model-consistent measure of estimation precision, especially for reporting or comparing to HLM7-style results. Use the alternative if you’re exploring the empirical variability of random effects in a specific sample and want a quick, descriptive ratio.

DominiqueMakowski:
@bwiernik what do you think of the following:

### Proposed Adjustment
The suggestion is to compute reliability as coefs[i]^2 / (coefs[i]^2 + SEs[i]^2) for each level (cluster) of the random effect, then average these values for each parameter (e.g., (Intercept) and Petal.Width). This:
- Uses squared coefficients (coefs[i]^2) and squared standard errors (SEs[i]^2) to work in variance-like units, consistent with hlmer.
- Calculates a reliability ratio per level, then averages across levels, mirroring hlmer’s cluster-level approach.

strengejacke:
> ### About hlmer
> ...

<sub>Gemini, please summarize the most important key messages</sub>

DominiqueMakowski:
I've ran a small simulation to compare the two reliability index computations (as a raw ratio or as an index of precision, `Reliability2`, between 0-1 - closer to hlmer)

<details>

<summary>Code</summary>


```r
library(tidyverse)
library(patchwork)
library(rstanarm)
options(mc.cores = parallel::detectCores()-4)

simulate_repeated <- function(n_ppt=40, n_trials=10,
                              between_sd=1,
                              between_intercept_sd=between_sd,
                              between_slope_sd=between_sd,
                              within_sd = 2,
                              fixed_intercept = 0,
                              fixed_slope = 1,
                              difficulty_range = c(0, 1)) {

  # Total number of observations
  n_total <- n_ppt * n_trials

  # Create participant IDs
  Participant <- factor(rep(paste0("P", 1:n_ppt), each = n_trials))

  # Simulate random effects for each participant
  set.seed(123)  # For reproducibility
  random_intercepts <- rnorm(n_ppt, mean = 0, sd = between_intercept_sd)
  random_slopes <- rnorm(n_ppt, mean = 0, sd = between_slope_sd)

  # Assign random effects to each observation
  random_intercepts_long <- rep(random_intercepts, each = n_trials)
  random_slopes_long <- rep(random_slopes, each = n_trials)

  # Simulate Difficulty (uniformly distributed within range)
  Difficulty <- runif(n_total, min = difficulty_range[1], max = difficulty_range[2])

  # Simulate residual (within-participant) noise
  residual_noise <- rnorm(n_total, mean = 0, sd = within_sd)

  # Compute Score: Fixed effects + Random effects + Noise
  Score <- (fixed_intercept + random_intercepts_long) +
    (fixed_slope + random_slopes_long) * Difficulty +
    residual_noise

  # Combine into a data frame
  data <- data.frame(
    Participant = Participant,
    Difficulty = Difficulty,
    Score = Score
  )

  data
}


visualize_repeated <- function(df) {
  ggplot(df, aes(x=Difficulty, y=Score)) +
    geom_point(aes(color=Participant), alpha=0.3) +
    geom_smooth(aes(color=Participant), method="lm", formula = 'y ~ x', se=FALSE) +
    geom_smooth(method="lm", formula = 'y ~ x', se=FALSE, color="black") +
    theme_minimal() +
    theme(legend.position = "none")
}


((visualize_repeated(simulate_repeated(n_ppt = 30, n_trials = 10, between_sd=1)) +
  labs(title = "n_ppt = 30, n_trials = 10, sd=1")) |
  (visualize_repeated(simulate_repeated(n_ppt = 30, n_trials = 100, between_sd=1)) +
  labs(title = "n_ppt = 30, n_trials = 100, sd=1"))) /
  ((visualize_repeated(simulate_repeated(n_ppt = 30, n_trials = 10, between_sd=4)) +
     labs(title = "n_ppt = 30, n_trials = 10, sd=4")) |
  (visualize_repeated(simulate_repeated(n_ppt = 30, n_trials = 100, between_sd=4)) +
     labs(title = "n_ppt = 30, n_trials = 100, sd=4")))


rez <- data.frame()
for(n_trials in seq(40, 250, by=20)) {
  print(n_trials)
  for(between_sd in c(0.5, 2, 4)) {
    df <- simulate_repeated(n_ppt = 30, n_trials = n_trials, between_sd=between_sd)
    rez1 <- lme4::lmer(Score ~ Difficulty + (Difficulty|Participant), data = df) |>
      performance::check_reliability() |>
      mutate(Model = "lme4")

    rez2 <- glmmTMB::glmmTMB(Score ~ Difficulty + (Difficulty|Participant), data = df) |>
      performance::check_reliability() |>
      mutate(Model = "glmmTMB")

    # rez3 <- rstanarm::stan_lmer(Score ~ Difficulty + (Difficulty|Participant), data = df, refresh=0) |>
    #   performance::check_reliability() |>
    #   mutate(Model = "rstanarm")

    rez4 <- brms::brm(Score ~ Difficulty + (Difficulty|Participant), data = df, refresh=0, backend="cmdstanr") |>
      performance::check_reliability() |>
      mutate(Model = "brms",
             Parameter = ifelse(Parameter == "Intercept", "(Intercept)", Parameter))

    rez_ <- rbind(rez1, rez2, rez4)
    rez_$n_trials <- n_trials
    rez_$between_sd <- between_sd

    rez <- rbind(rez, rez_)
  }
}


visualize_reliability <- function(rez, col="Reliability") {
  rez$Reliability <- rez[[col]]
  p <- rez |>
    ggplot(aes(x=n_trials, y=Reliability)) +
    geom_point(aes(color=Model)) +
    geom_smooth(aes(color=Model, fill=Model, linetype=as.factor(between_sd)), method = 'loess', formula = 'y ~ x', alpha=0.1) +
    facet_wrap(~Parameter, scales="free") +
    theme_minimal()
  if(col == "Reliability") p <- p + scale_y_log10()
  p
}

visualize_reliability(rez) / visualize_reliability(rez, col="Reliability2")

```

</details>

image

Some take-aways:
- Reliability increases as expected as a function of the number of trials as well as between-subject variability
- Reliability1 (raw ratio) and 2 display the same patterns: they seem pretty much equivalent
  - Pros: bounded index can be useful
  -  Cons: the raw ratio's meaning is more intuitive (how much more there is between variability relative to within variability)
- Differences from the same data between model: 
  - glmmTMB gives similar results to brms, which I consider to be the ground-truth 
  - `lme4` inflates reliability indices: suggesting that it underestimates within variability/overestimates between variability

DominiqueMakowski:
We have everything we need in `estimate_grouplevel()` I think

strengejacke:
Ok, wasn't sure if those are accurate indices (like taking the SD of the coefficients, instead of relying on the output of `VarCorr()`.

strengejacke:
``` r
data(sleepstudy, package = "lme4")
m <- glmmTMB::glmmTMB(Reaction ~ Days + (1 | Subject), data = sleepstudy)
#> Warning in check_dep_version(dep_pkg = "TMB"): package version mismatch: 
#> glmmTMB was built with TMB package version 1.9.16
#> Current TMB package version is 1.9.17
#> Please re-install glmmTMB from source or restore original 'TMB' package (see '?reinstalling' for more information)

sd(modelbased::estimate_grouplevel(m)$Coefficient)
#> [1] 35.76336
lme4::VarCorr(m)
#> 
#> Conditional model:
#>  Groups   Name        Std.Dev.
#>  Subject  (Intercept) 36.012  
#>  Residual             30.895
```

<sup>Created on 2025-03-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Can elaborate on 

> SE^2 (noise): The sampling variance, which depends on cluster size and residual variance.

How is SE^2 exactly computed?

R/check_reliability.R
```diff
@@ -0,0 +1,104 @@
...
+          Component = comp,
+          Group = grp,
+          Parameter = param,
+          Variability = stats::sd(d[[coefname]]),
+          Uncertainty = mean(d[[dispname]])
+        )
+        rez$Reliability <- rez$Variability / rez$Uncertainty
+
+        # Alternative: average of level-specific reliability
+        rez$Reliability2 <- mean(d[[coefname]]^2 / (d[[coefname]]^2 + d[[dispname]]^2))
```
DominiqueMakowski: > How is SE^2 exactly computed?

In the "empirical" approach (based on `estimate_grouplevel()`, rather than on variance parameters estimated by the model itself), like so lol

strengejacke: :thinking:

I indeed meant the approach in hlme.

strengejacke:
> Ok, wasn't sure if those are accurate indices (like taking the SD of the coefficients, instead of relying on the output of `VarCorr()`.

Ok, now I understand the differences in the nominator, based on your above writing.
I wonder if we can also have a similar denominator in the alternative formula, so the index also ranged from 0 to 1. I find it quite intuitive, especially since you else don't immediately know the bounds/limits of the index (if it can exceed 1).

DominiqueMakowski:
> compare with ML on lme4?

Let me try

strengejacke:
If we know how SE^2 is computed in hlmer, I suggest we pick a similar denominator. I added a third alternative, but not sure it makes more sense right now.

strengejacke:
I think this code for the third alternative now makes sense. tau = SD^2. so ratio would be SD^2 / (SD^2+SE^2)

R/check_reliability.R
```diff
@@ -0,0 +1,108 @@
...
+
+        # Alternative 2: like hlmer?
+        rez$Reliability3 <- rez$Variability^2 / (rez$Variability^2 + rez$Uncertainty^2)
```
DominiqueMakowski: Grok said:

The suggestion is to compute reliability as coefs[i]^2 / (coefs[i]^2 + SEs[i]^2) for each level (cluster) of the random effect, then average these values for each parameter (e.g., (Intercept) and Petal.Width). This:
- Uses squared coefficients (coefs[i]^2) and squared standard errors (SEs[i]^2) to work in variance-like units, consistent with hlmer.
- Calculates a reliability ratio per level, then averages across levels, mirroring hlmer’s cluster-level approach.

But indeed, I don't know whether it's better to reliability per level **then** average or average first 🤷

strengejacke: > Grok 

You're on the loser's road... 😬 

Given that the reliability index is the ratio of estimate variability to error variability, SD^2 seems to make more sense (see reliability3):

```r
m <- insight::download_model("brms_mixed_10")
x <- modelbased::estimate_grouplevel(m)

performance::check_reliability(m)
#> The number of random effects group levels (N=3) might be too low to
#>   reliably estimate the variability.
#>   Group   Parameter Variability Uncertainty Reliability Reliability2
#> 1 Group   Intercept  0.92950881   0.8448505    1.100205    0.3569732
#> 2 Group Petal.Width  0.04800641   0.1873084    0.256296    0.0467122
#>   Reliability3
#> 1   0.54760370
#> 2   0.06163875
```

strengejacke:
Looking at the paper (Eq. 1) and the related code:
https://github.com/specl/ctx-reliability/blob/8a7df0b373f8c4d6238fa581b532cd918625afd8/v3/p.Rmd#L111-L131

it looks like the reliability index is:

```r
v <- insight::get_variance(m)
L <- 10 # trial size

var_between <- v$var.intercept
var_within <- v$var.residual
gamma2 <- var_between / var_within
Er <- gamma2 / (gamma2 + (2 / L))
```

Only question is how to calculate / retrieve the trial size, `L`?

``` r
library(ggplot2)
library(glmmTMB)
#> Warning in check_dep_version(dep_pkg = "TMB"): package version mismatch: 
#> glmmTMB was built with TMB package version 1.9.16
#> Current TMB package version is 1.9.17
#> Please re-install glmmTMB from source or restore original 'TMB' package (see '?reinstalling' for more information)

data(sleepstudy, package = "lme4")
m <- glmmTMB(Reaction ~ Days + (1 + Days | Subject), data = sleepstudy)
x <- modelbased::estimate_grouplevel(m)

performance::check_reliability(m)
#>     Group   Parameter Variability Uncertainty Reliability Reliability2
#> 1 Subject (Intercept)   21.201768   13.407475    1.581339    0.4780055
#> 2 Subject        Days    5.398442    2.680198    2.014195    0.5144717
#>   Reliability3
#> 1    0.7143375
#> 2    0.8022535

v <- insight::get_variance(m)
var_between <- v$var.intercept
var_within <- v$var.residual
gamma2 <- var_between / var_within
Er <- function(L) gamma2 / (gamma2 + (2 / L))

d <- data.frame(
  L = 1:1000,
  Reliability = Er(1:1000)
)

ggplot(d, aes(x = L, y = Reliability)) +
  geom_line() +
  scale_x_log10() +
  xlab("Trial Size (L)") +
  ylab("Reliability Coefficient")
```



<sup>Created on 2025-03-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Using your code, you see the three reliability indices. Number 2 is your original implementation, number 3 is my addition, number 4 is the one the way I understood it from the paper. Not sure if this makes sense? What would be the true value?

<details>

<summary>Code</summary>

``` r
library(tidyverse)
library(patchwork)
library(rstanarm)

simulate_repeated <- function(n_ppt=40, n_trials=10,
                              between_sd=1,
                              between_intercept_sd=between_sd,
                              between_slope_sd=between_sd,
                              within_sd = 2,
                              fixed_intercept = 0,
                              fixed_slope = 1,
                              difficulty_range = c(0, 1)) {

  # Total number of observations
  n_total <- n_ppt * n_trials

  # Create participant IDs
  Participant <- factor(rep(paste0("P", 1:n_ppt), each = n_trials))

  # Simulate random effects for each participant
  set.seed(123)  # For reproducibility
  random_intercepts <- rnorm(n_ppt, mean = 0, sd = between_intercept_sd)
  random_slopes <- rnorm(n_ppt, mean = 0, sd = between_slope_sd)

  # Assign random effects to each observation
  random_intercepts_long <- rep(random_intercepts, each = n_trials)
  random_slopes_long <- rep(random_slopes, each = n_trials)

  # Simulate Difficulty (uniformly distributed within range)
  Difficulty <- runif(n_total, min = difficulty_range[1], max = difficulty_range[2])

  # Simulate residual (within-participant) noise
  residual_noise <- rnorm(n_total, mean = 0, sd = within_sd)

  # Compute Score: Fixed effects + Random effects + Noise
  Score <- (fixed_intercept + random_intercepts_long) +
    (fixed_slope + random_slopes_long) * Difficulty +
    residual_noise

  # Combine into a data frame
  data <- data.frame(
    Participant = Participant,
    Difficulty = Difficulty,
    Score = Score
  )

  data
}

Er <- function(L, gamma2) gamma2 / (gamma2 + (2 / L))

rez <- data.frame()
for(n_trials in seq(10, 250, by=30)) {
  print(n_trials)
  for(between_sd in c(0.5, 2, 4)) {
    df <- simulate_repeated(n_ppt = 40, n_trials = n_trials, between_sd=between_sd)

    mglmm <- glmmTMB::glmmTMB(Score ~ Difficulty + (Difficulty|Participant), data = df)
    rez2 <- performance::check_reliability(mglmm)

    v <- insight::get_variance(mglmm)
    var_between <- v$var.intercept
    var_within <- v$var.residual
    
    # rez_ <- rbind(rez1, rez2)
    rez_ <- data.frame(
      Reliability2 = mean(rez2$Reliability2),
      Reliability3 = mean(rez2$Reliability3),
      Reliability4 = Er(n_trials, var_between / var_within),
      n_trials = n_trials,
      between_sd = between_sd
    )
    rez <- rbind(rez, rez_)
  }
}

out <- datawizard::reshape_longer(
  rez,
  select = starts_with("Reliab"),
  names_to = "Type",
  values_to = "Value"
)

out$between_sd <- paste0("Between SD = ", out$between_sd)


ggplot(out, aes(x=n_trials, y=Value)) +
  geom_point(aes(color=Type)) +
  geom_line(aes(color=Type)) +
  facet_wrap(~between_sd) +
  theme_minimal()
```

</details>

strengejacke:
And here if you have a random-intercept-only model (dropping the random slope). Just for comparison, because we now don't need to "average" over random slopes _and_ intercepts.

strengejacke:
Not sure if this makes sense?

``` r
# Add groups to the data
data <- iris
data$Place <- as.factor(rep(c("P1", "P2", "P3", "P4", "P5", "P6"), each = 25))

# lme4
m <- lme4::lmer(Sepal.Width ~ Petal.Width + (Petal.Width | Place), data = data)

m |> 
  modelbased::estimate_grouplevel() |> 
  performance::check_reliability()
#>   Group   Parameter Variability Uncertainty Reliability Reliability2
#> 1 Place (Intercept)   0.9031420   0.2464442    3.664691    0.7711897
#> 2 Place Petal.Width   0.2146867   0.2110126    1.017411    0.3995945
#>   Reliability3 Reliability4
#> 1    0.9306997    0.9911697
#> 2    0.5086300    0.9911697


m <- lme4::lmer(Reaction ~ Days + (1 + Days | Subject), data = lme4::sleepstudy)

m |> 
  modelbased::estimate_grouplevel() |> 
  performance::check_reliability()
#>     Group   Parameter Variability Uncertainty Reliability Reliability2
#> 1 Subject (Intercept)   21.596038   12.070857    1.789106    0.5132405
#> 2 Subject        Days    5.455216    2.304839    2.366853    0.5571451
#>   Reliability3 Reliability4
#> 1    0.7619557    0.8237248
#> 2    0.8485306    0.8237248
```

<sup>Created on 2025-03-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
@emmalbenn @anafneves 

image

DominiqueMakowski:
> Not sure if this makes sense?

The fact that Reliability4 is the same for all random effects is probably not a good sign

strengejacke:
> The fact that Reliability4 is the same for all random effects is probably not a good sign

It's by design (it's just recycled from the data frame, actually it's only one value). The former indices calculate one variability value per random effects group, while the last index (from the paper, as far as I understood), calculates the "overall" variability of random effects (like we do in `insight::get_variance()`, and that's also what we use to calculate R2 or ICC).

So we might end up with two indices, one which calculates the reliability index for the model, and one for each "level", similar to the `by_group` argument we have in `icc()` and `r2_nakagawa()`. Though my understanding is that the reliability index is just the "overall" measure.

strengejacke:
from the docs of `insight::get_variance()`:
image

strengejacke:
btw, based on the pre-print, one of their main reliability indices is just the adjusted ICC we get from `performance::icc()`. 😬

DominiqueMakowski:
I think they answer two different questions:

1) **"How important are the random factors in the model?"** The part of variance withheld in the random grouping structure. Measured by ICC and your "general" reliability indices. This really has to do with general model performance, and is helpful to know whether to include some random parameters or not in the model (e.g., in the context of convergence issues, etc).

2) **"Are the between-groups differences measured reliably?"** Regardless whether the random factors are important, explain a large part of variance, are relevant etc., whether the individual-level variability is captured reliably (relative to noise / measurement error, i.e., within-individual-level variability).


The first question is useful for model performance & fit assessment, the second question is useful in-and-of-itself as an assessment of an (experimental) manipulation:
- Do I have enough children within each school so that my estimates of individual "schools" are meaningful?
- Do I have enough trials per participant so that my assessment of each individual's performance is reliable?
- Do my two conditions (placebo followed by controlled, within-subject design) yield a reliable difference at an individual level?

In other words, some indices are `performance_*` like indices (are about the model and its specification), and some other indices are checks that can be performed on a model but are about some stuff estimated by the model and answers questions related to the "data"

DominiqueMakowski:
I finally (tried to) read Rouder 2024, and I think we successfully implemented their index, **reliability**, as well as our alternative, with a temporary name of **D-vour** (1. inspired by `D'` *d-prime*, an index of signal-to-noise discrimination; 2. transparent: it is the Variability-Over-Uncertainty Ratio; and 3. Because it sounds cool and DEVOURs variance)

Those two might be somewhat distinct:
- Rouder's reliability relies on estimated model variances components, and is adjusted for the number of observations. To be honest, I'm not too sure what to make of it. 
- D-vour is an assessment of inter-level variability. It is straightforward to compute and interpret, and it increases with more observations (which is a feature not a bug)

It seems like the former is tied to "test-restest" reliability, while D-vour would be more about the reliability/robustness of inter-individual variability quantification

DominiqueMakowski:
And yes, maybe we should make it instead a `performance_*` function?... because arguably there is nothing to "check" (in the sense pass/fail), the values are what they are

strengejacke:
We might need additional methods for Bayesian models.

R/performance_reliability.R
```diff
@@ -0,0 +1,213 @@
...
+#' @details TODO: Add details.
+#'
+#' Interpretation: d-vour corresponds to: between-groups variability /
+#' (between-groups variability + within-group variability) A d-vour of 3/4
+#' (0.75) means that there is 3 times more variability between groups than
+#' within groups (3/1), and a d-vour of less than 0.5 means that there is more
+#' variability within groups than between groups (which is bad if the goal is to
+#' analyze group-level effects).
+#'
+#' @references TODO.
```
DominiqueMakowski: @emmaLbenn add the refs please

R/performance_reliability.R
```diff
@@ -1,22 +1,64 @@
...
+#'
+#' \deqn{\text{D-vour} = \frac{\sigma_B^2}{\sigma_B^2 + \mu_{\text{SE}}^2}}
+#'
+#' where:
+#' - \eqn{\sigma_B^2} is the **between-group variability** (computed as the SD of the random effect estimates).
+#' - \eqn{\mu_{\text{SE}}^2} is the **mean squared uncertainty** in random effect estimates (i.e., the average uncertainty).
+#'
+#'
+#' ### Interpretation:
+#' - **D-vour > 0.75**: Strong group-level effects (between-group variance is at least 3× greater than uncertainty).
```
strengejacke: The `3×` _might_ cause troubles rendering the PDF (not sure, but _can_ be), so we must remember this when submission to CRAN complains about invalid chars when creating the PDF manual.

strengejacke:
What's the state of this PR? Would probably submit an update of _performance_ in the next weeks...

strengejacke:
Ah, here:

> We might need additional methods for Bayesian models.

Tests are currently commented out, Bayesian models not yet working.

DominiqueMakowski:
Will get to it soon

Also: https://cogpsychreserve.netlify.app/posts/dotprobereliability/

DominiqueMakowski:
I think it's good to go! We have a note  for Bayesian models and voilà, it's a good first step!

PR #804, opened 2025-03-20
R2 incorrect for methods that use update(model, ~1) and have missing data

### LINKED ISSUE #803: R2 incorrect for methods that use update(model, ~1) and have missing data
Some R-squared calculations use the stats::update function to calculate the baseline model with just an intercept. See, for example, the [Nagelkerke R-squared](https://github.com/easystats/performance/blob/main/R/r2_nagelkerke.R) for `nnet::multinom`. 

```r
r2_nagelkerke.multinom <- function(model, ...) {
  l_base <- insight::get_loglikelihood(stats::update(model, ~1, trace = FALSE))
  .r2_nagelkerke(model, l_base)
}
```

An issue arises when the source dataframe has missing predictor values which lead to different sample sizes for the target model and the baseline model. Here's a quick demonstration confirming identical model fits but changing sample sizes for the baseline model.

```r
library(nnet)
library(performance)

n_obs <- 1000

softmax <- function(x){
  exp(x - max(x)) / sum(exp(x - max(x)))
}

sample_y <- function(x){
  sample(1:3, size = 1, prob = softmax(c(0.25 * x, -0.1 * x, 0 * x)))
}

sim_df <- data.frame(x = rnorm(n_obs, 0, 1), y = NA)

for(i in 1:nrow(sim_df)){
  sim_df$y[i] <- sample_y(sim_df$x[i])
}

sim_df$x[1:500] <- NA
sim_df2 <- sim_df[!is.na(sim_df$x),]

m1 <- multinom(y ~ x, data = sim_df)
m2 <- multinom(y ~ x, data = sim_df2)

r2_nagelkerke(m1) # Incorrect
r2_nagelkerke(m2) # Correct

c(nrow(m1$fitted.values), nrow(update(m1, ~1)$fitted.values))

c(nrow(m2$fitted.values), nrow(update(m2, ~1)$fitted.values))
```

and the output

```
> m1 <- multinom(y ~ x, data = sim_df)
# weights:  9 (4 variable)
initial  value 549.306144 
final  value 544.495584 
converged

> m2 <- multinom(y ~ x, data = sim_df2)
# weights:  9 (4 variable)
initial  value 549.306144 
final  value 544.495584 
converged


> r2_nagelkerke(m1) # Incorrect
Nagelkerke's R2 
      0.9018897 

> r2_nagelkerke(m2) # Correct
Nagelkerke's R2 
     0.01993749 

> c(nrow(m1$fitted.values), nrow(update(m1, ~1)$fitted.values))
[1]  500 1000

> c(nrow(m2$fitted.values), nrow(update(m2, ~1)$fitted.values))
[1] 500 500
```

strengejacke:
Fixes #803

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.13.0.1
+Version: 0.13.0.2
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -126,6 +126,7 @@ Suggests:
     multimode,
     nestedLogit,
     nlme,
+    nnet,
     nonnest2,
     ordinal,
     parallel,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -162,3 +162,4 @@ Config/Needs/website:
     r-lib/pkgdown,
     easystats/easystatstemplate
 Config/rcmdcheck/ignore-inconsequential-notes: true
+Remotes: easystats/insight
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,14 @@
   the full model), or can return singularity checks for each random effects term
   separately.
 
+## Bug fixes
+
+* Fixed issue with wrong computation of pseudo-R2 for some models where the
+  base-model (null model) was updated using the original data, which could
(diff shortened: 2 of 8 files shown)


strengejacke:
Tests in McFadden fail.
We need to extract data from env, not model frame, then filter for complete cases (only data for model predictors). passing model-frame to `update()` doesn't work with variables like `y / 10` etc.

PR #815, opened 2025-05-09
Revert nested change

mattansb:
@strengejacke we can keep working here, if you have more thoughts.

We can make two types of distinctions:
- Is variable X nested within groups (every group has a unique set of values of X)
- Is variable X crossed with groups (all groups have all possible values of X)
- (Or other)

But also:
- Does variable X vary between groups (is it correlated with group)
- Does variable X vary within groups
- (Or is correlated and varies within group)

So we have the following combinations (with "--" marking impossible situations):

|                | nested    | crossed  | other |
|----------------|-----------|----------|--------|
| **varies only between** | "between" | --       | "between"     |
| **varies only within**  | --  | "within" | --     |
| **both**           | "nested"        | --   | "both" |

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.13.0.9
+Version: 0.13.0.10
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/R/check_group_variation.R
+++ b/R/check_group_variation.R
@@ -186,19 +186,20 @@ check_group_variation.data.frame <- function(x,
 
   # create all combinations that should be checked
   combinations <- expand.grid(
-    variable = select,
-    group = by,
+    Variable = select,
+    Group = by,
     stringsAsFactors = FALSE
   )
-  combinations <- combinations[combinations$variable != combinations$group, ]
-  combinations$type <- NA_character_
+  combinations <- combinations[combinations$Variable != combinations$Group, ]
+  combinations$Variation <- NA_character_
+  combinations$Structure <- NA_character_
 
   # initialize lists
   for (i in seq_len(nrow(combinations))) {
-    combinations[i, "type"] <- .check_nested(
+    combinations[i, c("Variation", "Structure")] <- .check_nested(
       x,
-      combinations[i, "group"],
-      combinations[i, "variable"],
+      combinations[i, "Group"],
+      combinations[i, "Variable"],
       tolerance_numeric = tolerance_numeric,
       tolerance_factor = tolerance_factor
     )
@@ -206,8 +207,8 @@ check_group_variation.data.frame <- function(x,
 
   combinations <- datawizard::data_relocate(
     combinations,
-    select = "group",
-    before = "variable"
+    select = "Group",
+    before = "Variable"
   )
   class(combinations) <- c("check_group_variation", class(combinations))
   combinations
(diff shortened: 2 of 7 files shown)


strengejacke:
ok. but nesting and group variation are more or less two different things, or not?
I'm interested if a variable varies within or between a group. *Additionally*, the variable can be nested or not. Just having the information that something is "nested" doesn't tell me anything about the variation between or within groups.

My aim was to have a unique label for each combination:

|                | nested    | crossed  | other |
|----------------|-----------|----------|--------|
| **varies only between** | "between, nested" | --       | "between"     |
| **varies only within**  | --  | "within" | --     |
| **both**           | "both, nested"        | --   | "both" |

mattansb:
Maybe we can have an additional column that checks nested/crossed?

strengejacke:
> If you have a "between, nested" and "between (not nested)", you would probably treat those the same?

Yes, sure, but if the purpose is to inform about two characteristics, nesting and variation, then only reporting one can be misleading. I would think "oh, this is nested", and for the other case "why isn't it nested, even if lme4::is_nested() tells me so?".

We should either change labels, or add a column that indicates nested/crossed.

mattansb:
Okay, I've implemented a separate column (and renamed the "type" to Variation), such that:
- The "Structure" column is `NA` unless a nested or crossed structure is detected. Should something else be shown?
- Only relevant for non-numerical variables (that are only check for within/between variation)
- Given always (should this be an additional feature, that is off by default?)

``` r
library(performance)

data(npk)
check_group_variation(npk, by = "block")
#> Check block variation
#> 
#> variable | Variation | Structure
#> --------------------------------
#> N        |    within |   crossed
#> P        |    within |   crossed
#> K        |    within |   crossed
#> yield    |      both |

egsingle <- data.frame(
  schoolid = factor(rep(c("2020", "2820"), times = c(18, 6))),
  lowinc = rep(c(TRUE, FALSE), times = c(18, 6)),
  childid = factor(rep(
    c("288643371", "292020281", "292020361", "295341521"),
    each = 6
  )),
  female = rep(c(TRUE, FALSE), each = 12),
  year = rep(1:6, times = 4),
  math = c(
    -3.068, -1.13, -0.921, 0.463, 0.021, 2.035,
    -2.732, -2.097, -0.988, 0.227, 0.403, 1.623,
    -2.732, -1.898, -0.921, 0.587, 1.578, 2.3,
    -2.288, -2.162, -1.631, -1.555, -0.725, 0.097
  )
)

check_group_variation(egsingle, by = c("schoolid", "childid"), include_by = TRUE)
#> Check schoolid variation
#> 
#> variable | Variation | Structure
#> --------------------------------
#> childid  |      both |    nested
#> lowinc   |   between |    nested
#> female   |      both |          
#> year     |    within |          
#> math     |      both |          
#> 
#> Check childid variation
#> 
#> variable | Variation | Structure
#> --------------------------------
#> schoolid |   between |          
#> lowinc   |   between |          
#> female   |   between |          
#> year     |    within |          
#> math     |      both |
```

<sup>Created on 2025-05-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Nice!

> The "Structure" column is NA unless a nested or crossed structure is detected. Should something else be shown?

No, I think that's ok.

> Given always (should this be an additional feature, that is off by default?)

I have no strong preference here. I leave this decision to you :-)

mattansb:
Alright (:

I renamed the "Structure" column to "Design".

PR #830, opened 2025-06-25
specifying binomial data as proportion for glmmTMB confuses check_predictions

### LINKED ISSUE #829: specifying binomial data as proportion for glmmTMB confuses check_predictions
For `glmmTMB` fits (but not for the equivalent `glm` fits), `check_predictions` gives bogus results when binomial models are specified as `glmmTMB(prop ~ ..., weights = ...)` rather than `glmmTMB(cbind(success, failure) ~ ...)`. AFAICT the problem is not with `predict.glmmTMB`, which gives identical predictions in both cases ...

```r
data("cbpp", package = "lme4")
fit_prop <- glmmTMB(incidence/size ~ period + herd, weights = size, family = binomial, data = cbpp)
fit_cbind <- update(fit_prop, cbind(incidence, size-incidence) ~ ., weights = NULL)
## model log-likelihoods are not just all.equal(), but identical() ...
stopifnot(identical(logLik(fit_prop), logLik(fit_cbind)))
stopifnot(identical(predict(fit_prop), predict(fit_cbind)))
```

```r
library(performance)
check_predictions(fit_prop)
```

Image

```r
check_predictions(fit_cbind)
```

Image

strengejacke:
Fixes #829

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.14.0.6
+Version: 0.14.0.7
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,11 @@
 * Added following methods for `psych::fa()`, `psych::principal()`, `item_omega()`,
   `psych::omega()`, and `parameters::factor_analysis()`: `check_normality()`, `check_residuals()`, and `check_outliers()`, `model_performance()`.
 
+## Bug fixes
+
+* Fixed issue in `check_predictions()` for models with a response defined as
+  proportion.
+
 # performance 0.14.0
 
 ## Breaking Changes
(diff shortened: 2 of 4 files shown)


bbolker:
I think `"(weights)" %in% names(model.frame(fit_prop))` would be  a more reliable rubric for whether weights were specified for a binomial model: if I decided to pre-calculate the proportion and use `prop ~ ...` rather than `successes/total ~ ...`, I think the current test would fail ...

bbolker:
Hmm. If `family == "binomial"` and weights are present what else could it be? (I don't know what happens if you specify a two-column matrix as the response *and* give a `weights` argument to `glm` ...)

From `?glm`, at least:

> Non-‘NULL’ ‘weights’ can be used to indicate that different
     observations have different dispersions (with the values in
     ‘weights’ being inversely proportional to the dispersions); or
     equivalently, when the elements of ‘weights’ are positive integers
     w_i, that each response y_i is the mean of w_i unit-weight
     observations.  For a binomial GLM prior weights are used to give
     the number of trials when the response is the proportion of
     successes: they would rarely be used for a Poisson GLM.

PR #874, opened 2025-11-14
Fix check_model() plots for categorical predictors

### LINKED ISSUE #873: check_model() linearity & variance for categorical predictors
For models with a categorical predictor (3 levels here):

```r
star <- read.csv("https://drmankin.github.io/disc_stats/star.csv") |>
  dplyr::mutate(dplyr::across(dplyr::starts_with("star"), forcats::as_factor))

star_maths_lm <- lm(math2 ~ star2, data = star, na.action = na.exclude)

performance::check_model(star_maths_lm)
```

Image

```r
plot(star_maths_lm, which = 1)
```

Image

The variance and linearity plots could arguably be more informative. In particular the one to compare the variance across the different groups.

In a way the very large CI masks the variability in the data points. But I'm not s ure what's the best way to adress it:

- Add an option to simply drop the CI ribbon?
- Have bespoke plots for categorical predictors? 
- Alternatives?

ANAMASGARD:
Fixes #873

- The linearity and variance plots were hard to read when models had only 
categorical predictors - the confidence ribbons were huge and covered up 
the actual data pattern.

- Now check_model() detects when all predictors are categorical and hides 
the dots automatically. Makes the variance across groups way easier to see.

- Works with factor(x), as.factor(x), or variables that are already factors.
- If your model has both categorical and continuous predictors, it still 
shows dots like before.

- You can override this with show_dots = TRUE if you want the old behavior.

- Added tests, updated docs. Everything passes locally on my system .

### DIFF
--- a/.github/workflows/format-suggest.yaml
+++ b/.github/workflows/format-suggest.yaml
@@ -28,7 +28,7 @@ jobs:
       pull-requests: write
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 


ANAMASGARD:
Sir @bwiernik! Here's the visual comparison:

## BEFORE (with dots - old behavior)
plot_BEFORE_with_dots

**Issues:**
- All 2,161 data points stack at just 3 x-positions (one per category level)
- Creates dense visual clutter
- Hard to see the variance pattern across groups
- The smooth line gets lost in the point cloud

## AFTER (without dots - new auto-detected behavior)
plot_AFTER_without_dots


**Improvements:**
- Clean view of the smooth line
- Easy to compare variance across the 3 categorical groups
- Pattern is immediately visible
- Much more informative for diagnostic purposes

---

### Key Difference

For categorical predictors, dots don't spread along the x-axis - they stack vertically at discrete positions. Removing them reveals the actual variance pattern, making it easier to assess model assumptions.

This only affects models with **all categorical predictors**. Mixed models (categorical + continuous) keep dots visible.

### Reproducible Code

library(performance)

star <- read.csv("https://drmankin.github.io/disc_stats/star.csv")
star$star2 <- as.factor(star$star2)
model <- lm(math2 ~ star2, data = star, na.action = na.exclude)

# BEFORE: show_dots = TRUE
plot(check_model(model, show_dots = TRUE))

# AFTER: auto-detected (show_dots = FALSE)
plot(check_model(model))---

**Note:** As @strengejacke mentioned, we could hide CI bands instead. I'm happy to adjust the approach based on the team's preference!

PR #932, opened 2026-08-05
correlation for `check_group_variation()`

mattansb:
This PR adds an `Eta` column to the output of `check_group_variation()` which is a numeric effect size of the grouping variable's predictive association strength: when it is 0 the grouping variable carries no predictive information, when it is 1 the variable is perfectly predicted by the grouping variable. 
- For numeric variables it is the `sqrt(`[`icc`](https://easystats.github.io/performance/reference/icc.html)`)` - also know as $\eta$ (eta).
- For non-numeric variables it is a _non-symmetric_ version of [Cramer's _V_](https://easystats.github.io/effectsize/reference/phi.html).

``` r
performance::check_group_variation(
  mlmRev::egsingle,
  by = c("schoolid", "childid"),
  include_by = TRUE
)
#> Check schoolid variation
#> 
#> Variable | Variation |  Design |     r
#> --------------------------------------
#> childid  |      both |  nested |  .430
#> year     |      both |         |  .171
#> grade    |      both |         |  .185
#> math     |      both |         |  .358
#> retained |      both |         |  .203
#> female   |    within | crossed |  .183
#> black    |      both |         |  .824
#> hispanic |      both |         |  .692
#> size     |   between |         | 1.000
#> lowinc   |   between |         | 1.000
#> mobility |   between |         | 1.000
#> 
#> Check childid variation
#> 
#> Variable | Variation | Design |     r
#> -------------------------------------
#> schoolid |   between |        | 1.000
#> year     |      both |        |  .416
#> grade    |      both |        |  .429
#> math     |      both |        |  .713
#> retained |      both |        |  .478
#> female   |   between |        | 1.000
#> black    |   between |        | 1.000
#> hispanic |   between |        | 1.000
#> size     |   between |        | 1.000
#> lowinc   |   between |        | 1.000
#> mobility |   between |        | 1.000
```

<sup>Created on 2026-08-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.17.1.6
+Version: 0.17.1.7
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -103,6 +103,7 @@ Suggests:
     dbscan,
     DHARMa (>= 0.4.7),
     discovr,
+    effectsize,
     estimatr,
     fixest,
     flextable,
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Changes
 
+* `check_group_variation()` now returns a numeric effect size of the grouping
+  variable's predictive association strength.
+
 * Updated test-files to fix deprecated names.
 
 * `r2()` now also covers remaining families from package *glmmTMB* and defaults
--- a/R/check_group_variation.R
+++ b/R/check_group_variation.R
@@ -73,12 +73,14 @@
 #' indicates if and which predictors are possibly affected by heterogeneity
 #' bias.
 #'
-#' @return A data frame with Group, Variable, Variation, Design, and Eta columns.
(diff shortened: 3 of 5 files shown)


strengejacke:
Nice! Do we have a reference for eta?

DominiqueMakowski:
Nice one indeed, would its usage be more meaningful in the context of exploratory analysis? Or to to help (mixed) model specification?

mattansb:
I've renamed the column to a more generic "r" (see updated example above) - the docs still explain that for numeric variables it is the correlation ratio $\eta$ (with reference).

> Nice one indeed, would its usage be more meaningful in the context of exploratory analysis? Or to to help (mixed) model specification?

Ideally one already knows the design of their data and so know which variables are within/between/both, but this is useful for validating your assumptions (which is why it is in `{performance}`)! See examples where I use it in teaching group-mean-centering (`datawizard::demean()`) [here](https://github.com/mattansb/Hierarchical-Linear-Models-foR-Psychologists/blob/7f0b8a4238f6f8f934c3f5ada923a2d670a271b2/05%20within-person%20fluctuation%20models/within-person%20fluctuation%20models.R).

mattansb:
@strengejacke okay to merge?

strengejacke:
Yes

PR #934, opened 2026-08-07
Count parameters, not components, in the adjusted McFadden penalty

### LINKED ISSUE #933: Wrong McFadden's adjusted R2?
The McFadden adjusted R² returned by this package differs substantially from the value obtained with `DescTools`. This may be related to the following line, which evaluates to `1`, because `insight::find_parameters()` returns a list with a single named element ("`conditional`"), and the model terms are stored within that element.

https://github.com/easystats/performance/blob/3a27807d3245591680dd536ce7fbce97481b21bf/R/r2_mcfadden.R#L36

haomeng797-ship-it:
Fixes #933, and the diagnosis there is exactly right.

`.r2_mcfadden()` computed the penalty as `k <- length(insight::find_parameters(model))`. `find_parameters()` returns a list of components, so for a plain GLM this is `length(list(conditional = ...))` = 1, whatever the model. The adjustment was effectively `(logLik - 1) / logLik0` for every model, and the adjusted value barely moved away from the unadjusted one.

On the reprex from the issue, `glm(am ~ mpg + hp + wt, family = binomial)` on mtcars: the package reported an adjusted R2 of 0.751 with the true value being 0.612. After switching to `insight::n_parameters(model)` (k = 4 here), both values agree with `DescTools::PseudoR2(m, c("McFadden", "McFaddenAdj"))` to all printed digits (0.7972202 / 0.6121632), and the new regression test pins the closed form `1 - (logLik - k)/logLik0` directly.

One existing expectation had encoded the old behavior: the polr test in `test-r2_mcfadden.R` expected an adjusted R2 of 0.04597, which is the k = 1 value. With the model's 8 parameters counted (6 coefficients + 2 thresholds, verified by hand) the adjusted value is 0.04213, and the expectation is updated accordingly.

I checked the rest of the package for the same pattern: `r2_zeroinflated()`, `check_overdispersion()` and `performance_aicc()` all count parameters correctly (they use `[["conditional"]]` or `flatten = TRUE`), so this was the only site.

Version bumped to 0.17.1.7.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: performance
 Title: Assessment of Regression Models Performance
-Version: 0.17.1.6
+Version: 0.17.1.7
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,12 @@
 
 ## Bug fixes
 
+* `r2_mcfadden()` now uses the number of estimated parameters in the penalty
+  of the adjusted R2. Previously the component list returned by
+  `insight::find_parameters()` was counted instead of the parameters inside
+  it, so the penalty was almost always 1 and the adjusted value barely
+  differed from the unadjusted one (#933).
+
 * `check_collinearity()` now properly warns when the `vcov` matrix is rank
   deficient (#922).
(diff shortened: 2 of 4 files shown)


strengejacke:
Thanks a lot, looks good to me!

PR #940, opened 2026-08-24
Add POOR

strengejacke:


DominiqueMakowski:
that's interesting, i was recently thinking of something similar, is there a reference to that? Can't this be generalized to any type of coefs not just odds ratios?

strengejacke:
Look for merlo and logistic multilevel. One reference is https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0153778 (should add that to the function, too)

PR #947, opened 2026-09-09
Fix check_singularity.lme tolerance and dropped count-model hint

sims1253:
### check_singularity.lme applies the tolerance to the wrong expression

In `R/check_singularity.R`, the `lme` method closes `abs()` after the comparison:

```r
any(abs(stats::na.omit(as.numeric(diag(nlme::getVarCov(x)))) < tolerance))
```

`abs()` receives the logical vector from `< tolerance`, not the variance estimates. The effective test is `estimate < tolerance` on the raw diagonal, so a strongly negative estimate counts as singular whatever its magnitude. 

### check_heteroscedasticity drops the count-model hint

In `R/check_heteroscedasticity.R`, the default method builds a longer message for count models but never assigns it, so the hint never reaches the user. 

Found during my most recent [ry](https://github.com/sims1253/ry) audit.


strengejacke:
Thanks, looks good to me! Let me also request an additional AI review

strengejacke:
Thanks! Comments are minor and don't need to be addressed.

